PR #4, opened 2024-06-25
feat: v3 codec

zimpha:


src/L1/rollup/ScrollChain.sol
```diff
@@ -575,12 +663,151 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        if (_secondBlob != bytes32(0)) revert ErrorFoundMultipleBlobs();
+    }
+
     /// @dev Internal function to commit chunks with version 0
     /// @param _totalL1MessagesPoppedOverall The number of L1 messages popped before the list of chunks.
     /// @param _chunks The list of chunks to commit.
     /// @param _skippedL1MessageBitmap The bitmap indicates whether each L1 message is skipped or not.
     /// @return _batchDataHash The computed data hash for the list of chunks.
-    /// @return _totalL1MessagesPoppedInBatch The total number of L1 messages poped in this batch, including skipped one.
+    /// @return _totalL1MessagesPoppedInBatch The total number of L1 messages popped in this batch, including skipped one.
```
colinlyguo, line 810: since this pr tends to reduce gas in the meantime, come with a question, could we remove `maxNumTxInChunk` check in this contract?

colinlyguo: because updating this value may need a timelock.

zimpha: we don't limit `maxNumTxInChunk` in circuit any more?

colinlyguo: still limit `maxNumTxInChunk`, just wondering if this check can only be done in verifier.

zimpha: verifier will fail to verify and then we need to revert batch.
early stop in commit always better.

colinlyguo: makes sense. and the upgrades when increasing this value could be async.

src/L1/rollup/ScrollChain.sol
```diff
@@ -436,76 +506,94 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        if (_postStateRoot == bytes32(0)) revert ErrorStateRootIsZero();
+
+        // retrieve finalized state root and batch hash from storage
+        uint256 _finalizedBatchIndex = lastFinalizedBatchIndex;
+        bytes32 _prevStateRoot = finalizedStateRoots[_finalizedBatchIndex];
+        bytes32 _finalizedBatchHash = committedBatches[_finalizedBatchIndex];
+
+        // compute pending batch hash and verify
+        (uint256 batchPtr, bytes32 _batchHash, uint256 _batchIndex, ) = _loadBatchHeader(_batchHeader);
+        if (_batchIndex <= _finalizedBatchIndex) revert ErrorBatchIsAlreadyVerified();
```
colinlyguo, line 573: Is there any place to check if all batches before this bundle are finalized?

in the mock contract test (removing proof verification part), if rollup-relayer finalizes a bundle (e.g., containing batch 3), then previous batches (batch 2) cannot be finalized because the `lastFinalizedBatchIndex` is updated to 3.

zimpha: if a bundle containing batch 3 is finalized, we assume batch 2 is also finalized. So checking `lastFinalizedBatchIndex` is enough to check whether a batch is finalized or not.

src/L1/rollup/ScrollChain.sol
```diff
@@ -322,109 +312,189 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        _afterCommitBatch(_batchIndex, _batchHash);
+    }
+
+    /// @inheritdoc IScrollChain
+    ///
+    /// @dev This function will revert unless all V0/V1/V2 batches are finalized. This is because we start to
+    /// pop L1 messages in `commitBatchWithBlobProof` but not in `commitBatch`. We also introduce `finalizedQueueIndex`
+    /// in `L1MessageQueue`. If one of V0/V1/V2 batches not finalized, `L1MessageQueue.pendingQueueIndex` will not
+    /// match `parentBatchHeader.totalL1MessagePopped` and thus revert.
+    function commitBatchWithBlobProof(
```
icemelon, line 352: shall we call it `commitBatchDarwin`?

src/L1/rollup/ScrollChain.sol
```diff
@@ -322,109 +312,189 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        BatchHeaderV1Codec.storeBlobVersionedHash(batchPtr, _blobVersionedHash);
+        BatchHeaderV1Codec.storeParentBatchHash(batchPtr, _parentBatchHash);
+
+        uint256 lastBlockTimestamp;
+        {
+            bytes memory lastChunk = _chunks[_chunks.length - 1];
+            lastBlockTimestamp = ChunkCodecV1.getLastBlockTimestamp(lastChunk);
+        }
+        BatchHeaderV3Codec.storeLastBlockTimestamp(batchPtr, lastBlockTimestamp);
+        BatchHeaderV3Codec.storeBlobDataProof(batchPtr, _blobDataProof);
```
icemelon, line 416: This part of code is very prone to bugs because it reuses multiple batch header codec version to store the header fields. Why do you choose it this way? It's more clean and less prone to bugs if each batch header codec has a complete support for its own version.

src/L1/rollup/L1MessageQueue.sol
```diff
@@ -64,8 +64,11 @@ contract L1MessageQueue is OwnableUpgradeable, IL1MessageQueue {
     /// @dev The bitmap for skipped messages, where `skippedMessageBitmap[i]` keeps the bits from `[i*256, (i+1)*256)`.
     mapping(uint256 => uint256) private skippedMessageBitmap;
 
+    /// @inheritdoc IL1MessageQueue
+    uint256 public finalizedQueueIndexPlusOne;
```
icemelon, line 68: Can rename to `totalFinalizedL1Messages`

src/L1/rollup/MultipleVersionRollupVerifier.sol
```diff
@@ -123,7 +123,19 @@ contract MultipleVersionRollupVerifier is IRollupVerifier, Ownable {
     ) external view override {
         address _verifier = getVerifier(_version, _batchIndex);
 
-        IZkEvmVerifier(_verifier).verify(_aggrProof, _publicInputHash);
+        IZkEvmVerifierV1(_verifier).verify(_aggrProof, _publicInputHash);
+    }
+
+    /// @inheritdoc IRollupVerifier
+    function verifyAggregateProof(
```
icemelon, line 130: should we call it `verifyBundleProof`?

zimpha: make sense

src/L1/rollup/ScrollChain.sol
zimpha: I want to reuse the code before. But if reuse the code by calling `BatchHeaderV1Codec. storeBlobVersionedHash` in `BatchHeaderV3Codec. storeBlobVersionedHash` would increase the gas costs. That's why I manually call `BatchHeaderV1Codec. storeBlobVersionedHash` outside.

icemelon: I mean just simply copy the code from `BatchHeaderV1Codec` to `BatchHeaderV3Codec`

zimpha: copy the code will make code size larger, also we always the [same code](https://github.com/scroll-tech/scroll-contracts/blob/2ac4f3f7e090d7127db4b13b3627cb3ce2d762bc/src/L1/rollup/ScrollChain.sol#L542) to decode `version` from header. If we copy the code, it will also be confused why we always use `BatchHeaderV0Codec` code. So I decide to add comments and use the previous version codes, instead of copy and use `BatchHeaderV3Codec`.

src/L1/rollup/L1MessageQueue.sol
zimpha: rename to `nexrUnfinalizedQueueIndex`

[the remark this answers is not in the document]
zimpha: rename to `nexrUnfinalizedQueueIndex`

src/L1/rollup/ScrollChain.sol
icemelon: ok

src/L1/rollup/L1MessageQueue.sol
```diff
@@ -367,9 +373,53 @@ contract L1MessageQueue is OwnableUpgradeable, IL1MessageQueue {
...
+        onlyScrollChain
+    {
+        uint256 cachedFinalizedQueueIndexPlusOne = nextUnfinalizedQueueIndex;
+        if (_newFinalizedQueueIndexPlusOne == cachedFinalizedQueueIndexPlusOne) return;
+        require(_newFinalizedQueueIndexPlusOne > cachedFinalizedQueueIndexPlusOne, "finalized index too small");
+        require(_newFinalizedQueueIndexPlusOne <= pendingQueueIndex, "finalized index too large");
+
+        nextUnfinalizedQueueIndex = _newFinalizedQueueIndexPlusOne;
+        unchecked {
+            emit FinalizedDequeuedTransaction(_newFinalizedQueueIndexPlusOne - 1);
```
colinlyguo, line 416: Currently bridge-history-backend only monitors `DequeueTransaction` and marks the skipped l1 messages immediately.

For UX of refunding, bridge-history-backend should monitor this event and tell Frontend that it can enable refunds after this event is emitted, otherwise, the tx will be reverted. 

One feasible change (without changing the Frontend) IMO:
1. adding a "pre-skipped" status, update l1 messages by listening to DequeueTransaction.
2. update status as "skipped" after FinalizedDequeuedTransaction (l1 messages <= _newFinalizedQueueIndexPlusOne - 1).

Do you think this is a real concern, or there are some other simple walkarounds? @zzq0826 @zimpha

zimpha: The bridge-history-backend should also monitor `ResetDequeuedTransaction`, in case batch reverting.

The frontend can check the `queueIndex` with `nextUnfinalizedQueueIndex` to decide whether it can enable refund.

colinlyguo: got it. will also add tracking `ResetDequeuedTransaction`, I think the backend can add a new status called "pre-skipped", for the frontend:
1. if the status is "pre-skipped", which means the message is very likely to be skipped (unless we revert the batch), but the user cannot trigger a refund yet.
2. if the status is "skipped", which means the message is skipped and the user can trigger the refund.
In this way, frontend does not need extra codes to check if the refund is available, only leaving the notification consideration to @zzq0826.

colinlyguo: but yeah. the method using "pre-skipped" won't be forward-compatible. @zimpha's solution is more reasonable and forward-compatible.

hardhat-test/ZkEvmVerifierV2.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+        }
+        // verify ok
+        const tx = await chain.finalizeBundleWithProof(header, postStateRoot, withdrawRoot, proof);
+        const receipt = await tx.wait();
+        console.log("Gas Usage:", receipt?.gasUsed);
+      });
+    });
+  };
+
+  for (const version of ["v0.12.0-rc.2", "v0.12.0-rc.3"]) {
```
colinlyguo, line 152: `v0.12.0` is released.

src/L1/rollup/ScrollChain.sol
```diff
@@ -713,18 +907,22 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         uint256 _length;
         if (version == 0) {
```
colinlyguo, line 908: We keep this because `importGenesisBatch` uses it? raise this question because, in commitBatch, version 0 is deprecated.

src/L1/rollup/ScrollChain.sol
```diff
@@ -713,18 +907,22 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
-            _batchIndex = BatchHeaderV1Codec.getBatchIndex(batchPtr);
+        } else if (version >= 3) {
+            (batchPtr, _length) = BatchHeaderV3Codec.loadAndValidate(_batchHeader);
         }
+
+        // the code for compute batch hash is the same for V0, V1, V2, V3
+        // also the `_batchIndex` and `_totalL1MessagesPoppedOverall`.
+        _batchHash = BatchHeaderV0Codec.computeBatchHash(batchPtr, _length);
+        _batchIndex = BatchHeaderV0Codec.getBatchIndex(batchPtr);
+        _totalL1MessagesPoppedOverall = BatchHeaderV0Codec.getTotalL1MessagePopped(batchPtr);
```
colinlyguo, line 920: Very small nitpick, putting this below `revert ErrorIncorrectBatchHash();` could save some gas usage.

src/L1/rollup/ScrollChain.sol
```diff
@@ -575,12 +649,151 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        returns (
+            uint256 batchPtr,
+            bytes32 _batchHash,
+            uint256 _batchIndex
+        )
+    {
+        if (_postStateRoot == bytes32(0)) revert ErrorStateRootIsZero();
+
+        // compute batch hash and verify
+        (batchPtr, _batchHash, _batchIndex, ) = _loadBatchHeader(_batchHeader);
```
colinlyguo, line 702:
```suggestion
        (_, _, _batchIndex, ) = _loadBatchHeader(_batchHeader);
```

src/L1/rollup/ScrollChain.sol
```diff
@@ -575,12 +649,151 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        view
+        returns (
+            bytes32 _parentBatchHash,
+            uint256 _batchIndex,
+            uint256 _totalL1MessagesPoppedOverall
+        )
+    {
+        // check whether the batch is empty
+        if (_chunks.length == 0) revert ErrorBatchIsEmpty();
+        (, _parentBatchHash, _batchIndex, _totalL1MessagesPoppedOverall) = _loadBatchHeader(_parentBatchHeader);
```
colinlyguo, line 669: should we check if `_batchIndex` (parent batch) is committed?

colinlyguo: oh i see. checked inside `_loadBatchHeader`.

src/mocks/ScrollChainMockFinalize.sol
```diff
@@ -27,95 +27,52 @@ contract ScrollChainMockFinalize is ScrollChain {
...
-            );
-            // We verify that the point evaluation precompile call was successful by testing the latter 32 bytes of the
-            // response is equal to BLS_MODULUS as defined in https://eips.ethereum.org/EIPS/eip-4844#point-evaluation-precompile
-            if (!success) revert ErrorCallPointEvaluationPrecompileFailed();
-            (, uint256 result) = abi.decode(data, (uint256, uint256));
-            if (result != BLS_MODULUS) revert ErrorUnexpectedPointEvaluationPrecompileOutput();
-        }
+        // compute pending batch hash and verify
+        (, bytes32 _batchHash, uint256 _batchIndex, ) = _loadBatchHeader(_batchHeader);
+        if (_batchIndex <= lastFinalizedBatchIndex) revert ErrorBatchIsAlreadyVerified();
```
colinlyguo, line 68: We should also check if the last batch is committed.

Because during adding unit tests, when the data are ready in db, and rollup-relayer calls `finalizeBundle` before committing these batches, the finalized height in the contract would be directly updated to `_batchIndex`, thus some batches cannot be committed to the contract.

zimpha: It is also checked in `_loadBatchHeader`.

src/L1/rollup/ScrollChain.sol
zimpha: Yes

src/mocks/ScrollChainMockFinalize.sol
```diff
@@ -27,95 +27,52 @@ contract ScrollChainMockFinalize is ScrollChain {
...
         withdrawRoots[_batchIndex] = _withdrawRoot;
 
-        // Pop finalized and non-skipped message from L1MessageQueue.
-        _popL1Messages(
-            BatchHeaderV1Codec.getSkippedBitmapPtr(memPtr),
-            BatchHeaderV1Codec.getTotalL1MessagePopped(memPtr),
-            BatchHeaderV1Codec.getL1MessagePopped(memPtr)
-        );
-
         emit FinalizeBatch(_batchIndex, _batchHash, _postStateRoot, _withdrawRoot);
```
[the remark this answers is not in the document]
zimpha: yes, we should. will add this later.

src/L1/rollup/ScrollChain.sol
zimpha: these two are returned and will be used by the caller.

src/mocks/ScrollChainMockFinalize.sol
colinlyguo: got it. after revisiting, I found the real issue is that if rollup-relayer calls finalizeBundle (without verifier), it will finalize <= batch index batches (including unfinalized codecv2 batches), if some codecv2 batches are not finalized yet, rollup-relayer will keep complaining that it cannot finalize these batches.

src/L1/rollup/ScrollChain.sol
colinlyguo: got it.

src/mocks/ScrollChainMockFinalize.sol
```diff
@@ -27,94 +27,56 @@ contract ScrollChainMockFinalize is ScrollChain {
...
+        );
 
-        // record state root and withdraw root
-        finalizedStateRoots[_batchIndex] = _postStateRoot;
-        withdrawRoots[_batchIndex] = _withdrawRoot;
+        // verify blob versioned hash
+        bytes32 _blobVersionedHash = BatchHeaderV1Codec.getBlobVersionedHash(batchPtr);
+        _checkBlobVersionedHash(_blobVersionedHash, _blobDataProof);
 
         // Pop finalized and non-skipped message from L1MessageQueue.
```
colinlyguo, line 47:
```suggestion
        // Pop finalized and non-skipped messages from L1MessageQueue.
```

src/mocks/ScrollChainMockFinalize.sol
```diff
@@ -27,94 +27,56 @@ contract ScrollChainMockFinalize is ScrollChain {
...
+        if (_batchIndex <= lastFinalizedBatchIndex) revert ErrorBatchIsAlreadyVerified();
 
-        // record state root and withdraw root
+        // store in state
+        // @note we do not store intermediate finalized roots
+        lastFinalizedBatchIndex = _batchIndex;
         finalizedStateRoots[_batchIndex] = _postStateRoot;
         withdrawRoots[_batchIndex] = _withdrawRoot;
 
         // Pop finalized and non-skipped message from L1MessageQueue.
```
colinlyguo, line 78:
```suggestion
        // Pop finalized and non-skipped messages from L1MessageQueue.
```

colinlyguo (approved):

PR #5, opened 2024-07-09
feat: deterministic deployment scripts

Thegaram:
Migrated from https://github.com/scroll-tech/scroll/pull/1303.

**Problem**: The current deployment scripts are slow and error-prone. We need to predict the deployment addresses of some contracts (e.g. for preallocating funds to `L2ScrollMessenger` in L2 genesis), but currently the addresses depend on the deployer private key and the order of deployment. If one transaction fails, subsequent transactions will lead to different addresses.

**Solution**:
- Use https://github.com/Arachnid/deterministic-deployment-proxy to make contract addresses much more predictable.
- Use a single Foundry script for deployment and initialization on L1 and L2 (basically merge the previous 5 scripts).
- Move genesis generation into the Foundry script. Previously this was a complex combination of Javascript, Rush, and bash scripts.
- Provide convenient Docker images. This makes deployment much more robust (we enforce a fixed Foundry version known to be working), and fast (compilation done in the build phase, no need to recompile when running the container).
- Use a TOML configuration file.

**Caveats**:
- Some of the deployed addresses still depend on the deployer account. The reason is that some contracts take the `owner` as a constructor argument, and the initial owner is the deployer account, so changing this also changes the `create2` deterministic address. But the addresses do not depend on deployment order anymore.


scripts/deterministic/DeployScroll.s.sol
```diff
@@ -0,0 +1,1272 @@
...
+            type(EmptyContract).creationCode
+        );
+    }
+
+    function deployL1Whitelist() private {
+        bytes memory args = abi.encode(DEPLOYER_ADDR);
+        L1_WHITELIST_ADDR = deploy("L1_WHITELIST", type(Whitelist).creationCode, args);
+    }
+
+    function deployL2GasPriceOracle() private {
```
zimpha, line 381: it is not used if we deploy `L1MessageQueueWithGasPriceOracle`.

docker/scripts/deploy.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "L1_RPC_ENDPOINT is not set"
+    L1_RPC_ENDPOINT="http://host.docker.internal:8543"
+fi
+
+if [ "$L2_RPC_ENDPOINT" = "" ]; then
+    echo "L2_RPC_ENDPOINT is not set"
+    L2_RPC_ENDPOINT="http://host.docker.internal:8545"
+fi
+
+if [ "${L1_RPC_ENDPOINT}" = "" ]; then
```
yiweichi, line 16: This code is repeated with line 6?

scripts/deterministic/DeployScroll.s.sol
```diff
@@ -0,0 +1,1622 @@
...
+            args
+        );
+
+        upgrade(L1_PROXY_ADMIN_ADDR, L1_GAS_TOKEN_GATEWAY_PROXY_ADDR, L1_GAS_TOKEN_GATEWAY_IMPLEMENTATION_ADDR);
+    }
+
+    function deployL1WrappedTokenGateway() private gasToken(true) {
+        bytes memory args = abi.encode(notnull(L1_WETH_ADDR), notnull(L1_STANDARD_ERC20_GATEWAY_PROXY_ADDR));
+
+        L1_WRAPPED_TOKEN_GATEWAY_ADDR = deploy(
```
Thegaram, line 1054: Is it intentional that this is not an upgradable proxy contract?

yiweichi: I think so, but needs @zimpha to confirm.

Thegaram: It's OK according to Xi.

scripts/deterministic/DeployScroll.s.sol
yiweichi: Raised a pr here to resolve this: https://github.com/scroll-tech/scroll-contracts/pull/38

yiweichi:
moved it here: https://github.com/scroll-tech/scroll-contracts/pull/63

PR #9, opened 2024-07-18
fix: make env.frontend fields generate from config

yiweichi:
Problem:
Fields like "REACT_APP_EXTERNAL_RPC_URI_L1" in .env.frontend file are hardcoded, not configurable.

Solution:
Modify scripts/deterministic/GenerateConfigs.s.sol to enable read those fields from config.toml file.

### DIFF
--- a/scripts/deterministic/Configuration.sol
+++ b/scripts/deterministic/Configuration.sol
@@ -74,6 +74,14 @@ abstract contract Configuration is Script {
     // coordinator
     string internal COORDINATOR_JWT_SECRET_KEY;
 
+    // frontend
+    string internal REACT_APP_EXTERNAL_RPC_URI_L1;
+    string internal REACT_APP_EXTERNAL_RPC_URI_L2;
+    string internal REACT_APP_BRIDGE_API_URI;
+    string internal REACT_APP_ROLLUPSCAN_API_URI;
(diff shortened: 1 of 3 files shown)


scripts/deterministic/Configuration.sol
```diff
@@ -74,6 +74,14 @@ abstract contract Configuration is Script {
     // coordinator
     string internal COORDINATOR_JWT_SECRET_KEY;
 
+    // frontend
```
Thegaram, line 77: Can you please also update `config-example.toml`?

scripts/deterministic/Configuration.sol
```diff
@@ -133,6 +141,13 @@ abstract contract Configuration is Script {
 
         COORDINATOR_JWT_SECRET_KEY = cfg.readString(".coordinator.COORDINATOR_JWT_SECRET_KEY");
 
+        REACT_APP_EXTERNAL_RPC_URI_L1 = cfg.readString(".frontend.REACT_APP_EXTERNAL_RPC_URI_L1");
```
Thegaram, line 144:
```suggestion
        RPC_URI_L1 = cfg.readString(".frontend.RPC_URI_L1");
```

Let's drop the `REACT_APP_` prefix from our config.toml file

Thegaram (approved):

PR #11, opened 2024-07-22
feat: add contracts for alternative gas token design

zimpha:


src/alternative-gas-token/L1ScrollMessengerNonETH.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**********************
+     * Internal Functions *
+     **********************/
+
+    /// @inheritdoc L1ScrollMessenger
+    function _sendMessage(
+        address _to,
+        uint256 _value,
```
Thegaram, line 52: I think it's a bit confusing to read this contract. Maybe easier to understand if we rename like this.
```suggestion
        uint256 _l2GasTokenValue,
```

Thegaram (commented):
Overall looks good

src/alternative-gas-token/L1ScrollMessengerNonETH.sol
```diff
@@ -0,0 +1,187 @@
...
+        bytes memory _message,
+        uint256 _gasLimit,
+        address _refundAddress
+    ) internal override {
+        // if we want to pass value to L2, must call from `L1NativeTokenGateway`.
+        if (_value > 0 && _msgSender() != nativeTokenGateway) revert ErrorNonZeroValueFromCaller();
```
Thegaram, line 58: Will it be possible to use the batch bridge gateway with `nativeTokenGateway`?

zimpha: no, need to modify the `L1BatchBridgeGateway` and `L2BatchBridgeGateway` to support.

src/alternative-gas-token/L1GasTokenGateway.sol
```diff
@@ -0,0 +1,186 @@
...
+        address _from,
+        address _to,
+        uint256 _amount,
+        bytes calldata _data
+    ) external payable override onlyCallByCounterpart nonReentrant {
+        if (msg.value > 0) {
+            revert ErrorNonZeroMsgValue();
+        }
+
+        uint256 downScaledAmount = _amount / scale;
```
Thegaram, line 117: I'm slightly concerned about this downscaling.

What if `l1_decimals = 6`, but I withdraw `N + 1/1e6 / 2` tokens 2 times?

Then we'd have `1/1e6` less tokens on L2 compared to L1, i.e. we'd have some dust (`1/1e6` tokens) stuck in the L1 gateway forever.

Maybe not an issue, just mentioning.

yiweichi (approved):

src/alternative-gas-token/L1GasTokenGateway.sol
Thegaram: Conclusion: it is ok to left some dust tokens. Users usually won’t care about this. and the invariant that "locked on L1 >= released on L2" still holds.

Thegaram:
fyi this is already merged into #5. we can merge into `main` once we decided on an audit plan. (This way we can ensure that these changes will not interfere with other changes pushed for mainnet.

Thegaram:
Closing as not planned.

PR #12, opened 2024-07-23
feat: log broadcast files

yiweichi:
Problem:
Some of clients want log files of the contracts that were successfully deployed, so they could get to know the contracts deployment process in detail.

Solution:
Foundry do provide a file -- broadcast/DeployScroll.s.sol/$CHAIN_ID_L1/run-latest.json to cache deployment information.
It doable to persist the file to a node, but kubernetes is supposed to be a cluster of machines, so it will be rather complex to get it.
**So here we cat broadcast files in docker, this would be easy to implement, and client shall get those information in deploy-contracts pod logs.**

### DIFF
--- a/docker/scripts/deploy.sh
+++ b/docker/scripts/deploy.sh
@@ -22,6 +22,14 @@ if [ "${BATCH_SIZE}" = "" ]; then
     BATCH_SIZE="100"
 fi
 
+if [ "${CHAIN_ID_L1}" = "" ]; then
+    CHAIN_ID_L1="111111"
+fi
+
+if [ "${CHAIN_ID_L2}" = "" ]; then
+    CHAIN_ID_L2="221122"
+fi
+
 echo "using L1_RPC_ENDPOINT = $L1_RPC_ENDPOINT"
 echo "using L2_RPC_ENDPOINT = $L2_RPC_ENDPOINT"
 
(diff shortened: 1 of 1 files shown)


docker/scripts/deploy.sh
```diff
@@ -22,6 +22,14 @@ if [ "${BATCH_SIZE}" = "" ]; then
     BATCH_SIZE="100"
 fi
 
+if [ "${CHAIN_ID_L1}" = "" ]; then
```
Thegaram, line 25: Let's not use defaults here. We can error if it is not provided.

In the `broadcast/...` path maybe you can also match using `*` since most likely there will only be 2 subfolders.

yiweichi: Yes, but it seems hard to determinate which one is L1 broadcast file and which one is L2 broadcast file.

yiweichi: btw, environment variable CHAIN_ID_L1 and CHAIN_ID_L1 are set by helm chart now when launching the container.

Thegaram: Can we do something like this?

```
echo ""
echo "Broadcast files:"
cat broadcast/DeployScroll.s.sol/*/run-latest.json
```

Or if it doesn't work we can use `CHAIN_ID_L1` and fail if it is not set.

dghelm:
Morty wants to close this for now, as we'll be achieving the logs through a more docker-native way, now that this isn't part of the k8s cluster.

PR #15, opened 2024-07-24
fix: log addresses script mode

yiweichi:
Fix function setScriptMode() in ./scripts/deterministic/DeterministicDeployment.sol set ScriptMode to WriteConfig when pass "log-addresses" as paremeter.


dghelm:
Fix looks right, but I think I'm missing context here. When will we expect to pass this parameter? The gen-config and deploy scripts that docker calls use the other modes -- is this just for being able to get the contract addresses without regenerating configuration files (and without re-deploying)?

yiweichi:
> Fix looks right, but I think I'm missing context here. When will we expect to pass this parameter? The gen-config and deploy scripts that docker calls use the other modes -- is this just for being able to get the contract addresses without regenerating configuration files (and without re-deploying)?

Yes, I think so.

PR #20, opened 2024-08-03
Feat deployment alt gas token

yiweichi:
Add alternative gas token contracts and deployment scripts.


zimpha (approved):

src/alternative-gas-token/L1WrappedTokenGateway.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    /// @dev Thrown when someone try to send ETH to this contract.
+    error ErrorCallNotFromFeeRefund();
+
+    /*************
+     * Constants *
+     *************/
+
+    /// @dev The safe gas limit used to bridge WETH to L2.
+    uint256 private constant SAFE_GAS_LIMIT = 200000;
```
zimpha, line 27: May need to increase this value to `450000`. If the first WETH deposit goes through this contract, we need at least `400000` to successfully create WETH token in L2.

If the first WETH deposit goes through gateway router, it is ok to use `200000` in this contract.

yiweichi: Yes, testing, and just realized it.

docker/templates/bridge-history-config.json
```diff
@@ -18,7 +18,9 @@
     "USDCGatewayAddr": "0x0000000000000000000000000000000000000000",
     "LIDOGatewayAddr": "0x0000000000000000000000000000000000000000",
     "DAIGatewayAddr": "0x0000000000000000000000000000000000000000",
-    "PufferGatewayAddr": "0x0000000000000000000000000000000000000000"
+    "PufferGatewayAddr": "0x0000000000000000000000000000000000000000",
+		"GasTokenGatewayAddr": null,
+    "WrappedTokenGatewayAddr": null
```
colinlyguo, line 23:
```suggestion
    "PufferGatewayAddr": "0x0000000000000000000000000000000000000000",
    "GasTokenGatewayAddr": null,
    "WrappedTokenGatewayAddr": null
```

scripts/deterministic/DeployScroll.s.sol
```diff
@@ -611,6 +640,38 @@ contract DeployScroll is DeterminsticDeployment {
...
+
+        bytes memory args = abi.encode(
+            "ScrollGasToken", // _name
+            "GasToken", // _symbol
+            decimal, // _decimals
+            DEPLOYER_ADDR, // _recipient
+            10**28 // _amount
+        );
+
+        L1_GAS_TOKEN_ADDR = deploy("L1_GAS_TOKEN", type(GasTokenExample).creationCode, args);
```
Thegaram, line 658: So if the dev does not specify the `L1_GAS_TOKEN` in `config.toml`, this will automatically deploy a test gas token contract, correct?

scripts/deterministic/DeployScroll.s.sol
```diff
@@ -933,8 +1036,14 @@ contract DeployScroll is DeterminsticDeployment {
     }
 
     function deployL2ETHGateway() private {
+        address COUNTERPART;
+        if (ALTERNATIVE_GAS_TOKEN_ENABLED) {
```
Thegaram, line 1040: Why do we still need to deploy `L2ETHGateway`? Doesn't L1 ETH map to a standard ERC20 token on L2?

docker/templates/chain-monitor-config.json
```diff
@@ -14,17 +14,19 @@
...
+          "gas_token_gateway": null
         },
         "scroll_messenger": null,
         "message_queue": null,
-        "scroll_chain": null
+        "scroll_chain": null,
+        "gas_token": null
     },
-    "start_messenger_balance": null
+    "start_messenger_balance": 0
```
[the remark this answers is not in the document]
Thegaram: I just realized it's missing. Added a draft here: https://github.com/scroll-tech/scroll-contracts/pull/24 probably needs some adjustments.

scripts/deterministic/GenerateConfigs.s.sol
```diff
@@ -128,10 +128,11 @@ contract GenerateChainMonitorConfig is DeployScroll {
         vm.writeJson(vm.toString(L1_CUSTOM_ERC20_GATEWAY_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.l1_gateways.custom_erc20_gateway");
         vm.writeJson(vm.toString(L1_ERC721_GATEWAY_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.l1_gateways.erc721_gateway");
         vm.writeJson(vm.toString(L1_ERC1155_GATEWAY_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.l1_gateways.erc1155_gateway");
+        vm.writeJson(vm.toString(L1_GAS_TOKEN_GATEWAY_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.l1_gateways.gas_token_gateway");
         vm.writeJson(vm.toString(L1_SCROLL_MESSENGER_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.scroll_messenger");
         vm.writeJson(vm.toString(L1_MESSAGE_QUEUE_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.message_queue");
         vm.writeJson(vm.toString(L1_SCROLL_CHAIN_PROXY_ADDR), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.l1_contracts.scroll_chain");
-        vm.writeJson(vm.toString(L2_DEPLOYER_INITIAL_BALANCE), CHAIN_MONITOR_CONFIG_PATH, ".l1_config.start_messenger_balance");
```
Thegaram, line 134: Should be `L2_DEPLOYER_INITIAL_BALANCE` if we use `ETH` and `0` if we use an alternative gas token.

docker/templates/chain-monitor-config.json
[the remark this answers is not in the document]
georgehao: I think contracts don't need chain_monitor config,  also the bridge-history-config.json

docker/templates/rollup-config.json
```diff
@@ -49,7 +49,7 @@
         "gas_price_diff": 50000
       },
       "chain_monitor": {
-        "enabled": false,
+        "enabled": true,
         "timeout": 3,
```
georgehao, line 53: Why contract need `bridge-history-api`, `rollup`, `chain-monitor` config

scripts/deterministic/DeployScroll.s.sol
```diff
@@ -186,6 +196,14 @@ contract DeployScroll is DeterminsticDeployment {
         _;
     }
 
+    /// @dev Only execute block if it's requied by alternative gas token mode.
+    modifier gasToken(bool gasTokenRequire) {
+        if (ALTERNATIVE_GAS_TOKEN_ENABLED != gasTokenRequire) {
```
georgehao, line 201: weird, IMO, if set `ALTERNATIVE_GAS_TOKEN_ENABLED=true` represents `alternative gas token` enabled, why need pass `gasTokenRequire` to check the consistency. if `ALTERNATIVE_GAS_TOKEN_ENABLED=fase && gasTokenRequire == false`, this check `(ALTERNATIVE_GAS_TOKEN_ENABLED != gasTokenRequire)` will also verify pass.

docker/templates/rollup-config.json
yiweichi: This is mainly because we want to have a single command to generate all config files for scroll-sdk, so here the file https://github.com/scroll-tech/scroll-contracts/blob/feat-deployment-alt-gas-token/docker/scripts/gen-configs.sh will help to do so.

docker/config-example.toml
```diff
@@ -13,6 +13,9 @@ MAX_L1_MESSAGE_GAS_LIMIT = 10000000
 
 L1_CONTRACT_DEPLOYMENT_BLOCK = 0
 
+ALTERNATIVE_GAS_TOKEN_ENABLED = true
+# EXAMPLE_GAS_TOKEN_DECIMAL = 6
```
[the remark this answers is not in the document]
yiweichi: Yes, that's for testing purpose, users should set their own gas token address (L1_GAS_TOKEN) on config.toml file.

scripts/deterministic/DeployScroll.s.sol
yiweichi: If alternative gas token enabled, the counterpart of `L1GasTokenGateway` is `L2ETHGateway`.
if alternative gas token disabled,  the counterpart of `L1ETHGateway` is `L2ETHGateway`.
So we always need L2ETHGateway.

docker/templates/chain-monitor-config.json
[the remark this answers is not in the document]
yiweichi: > I think contracts don't need chain_monitor config, also the bridge-history-config.json

Yes, this is mainly because we want to have a single command to generate all config files for scroll-sdk, so here the file https://github.com/scroll-tech/scroll-contracts/blob/feat-deployment-alt-gas-token/docker/scripts/gen-configs.sh will help to do so.

scripts/deterministic/DeployScroll.s.sol
yiweichi: Yes, that's true.

docker/templates/chain-monitor-config.json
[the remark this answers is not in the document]
georgehao: i see. but i don't like this way. why not generate the configs in scroll-sdk repo?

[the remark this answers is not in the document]
Thegaram: Yes, this is a convenient way to generate the configs before actually deploying the contracts. But I'm fine with moving the `GenerateConfigs` logic into `scroll-sdk`, or maybe add it to Daniel's cli.

scripts/deterministic/DeployScroll.s.sol
Thegaram: Some functions are only run when alternative gas token is enabled: `gasToken(true)`. Others are run when alternative gas token is disabled: `gasToken(false)`.

You need both, `ALTERNATIVE_GAS_TOKEN_ENABLED` is the user config for the rollup, while `gasTokenRequire` is for the script to decide to enable/disable certain functions.

If both are false then the check will evaluate to false (`false != false` is false), i.e. we will run the function instead of returning early.

docker/templates/chain-monitor-config.json
```diff
@@ -1,7 +1,7 @@
 {
   "l1_config": {
     "l1_url": null,
-    "confirm": "0x20",
+    "confirm": "0x0",
```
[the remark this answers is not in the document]
yiweichi: Yes, don't want to wait to long. I will change it back.

docker/templates/chain-monitor-config.json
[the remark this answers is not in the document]
yiweichi: > Why 0? L1 messenger balance will not be 0 in a real deployment.

For this, my original thinking is to burn rest of the deployer balance to all zero address, so we can set chain monitor `start_messenger_balance` to 0. but in this way L2 contracts deployment gas fee may not be burned.

but with https://github.com/scroll-tech/scroll-contracts/pull/24 , would be better.

scripts/deterministic/GenerateConfigs.s.sol
yiweichi: Chain monitor now will check erc20 gas token balance of `GastokenGateway` if we use an alternative gas token. the `start_messenger_balance ` in this case would represent the gas token balance of `GastokenGateway`.
So I think the `start_messenger_balance ` should always be L2_DEPLOYER_INITIAL_BALANCE.

PR #24, opened 2024-08-15
feat: lock tokens in bridge after deployment

Thegaram:


Thegaram:
fyi @yiweichi this is not a must-have for testnet, but it's a must-have for mainnet.

yiweichi:
made some adjustments, tested.

PR #33, opened 2024-09-03
fix: script check balance

yiweichi:
There is a issue when we check deployer balance, we can't get the real balance of scroll messenger or gas token gateway, because the `L1_SCROLL_MESSENGER_PROXY_ADDR` or `L1_GAS_TOKEN_GATEWAY_PROXY_ADDR` is still all zero address in that context.
And we shouldn't simulate the deployment of L1GasTokenGateway on layer2, because the deployment requires the decimal of the gas token.

### DIFF
--- a/scripts/deterministic/DeployScroll.s.sol
+++ b/scripts/deterministic/DeployScroll.s.sol
@@ -197,6 +197,14 @@ contract DeployScroll is DeterminsticDeployment {
         _;
     }
 
+    /// @dev Not execute block if we run the script on the specified layer.
+    modifier except(Layer layer) {
+        if (broadcastLayer == layer) {
+            return;
+        }
+        _;
+    }
+
     /// @dev Only execute block if it's requied by alternative gas token mode.
     modifier gasToken(bool gasTokenRequire) {
         if (ALTERNATIVE_GAS_TOKEN_ENABLED != gasTokenRequire) {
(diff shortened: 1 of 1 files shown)


scripts/deterministic/DeployScroll.s.sol
```diff
@@ -264,7 +272,8 @@ contract DeployScroll is DeterminsticDeployment {
 
         // check funds for initial deposit (L1, ETH as gas token)
         if (broadcastLayer == Layer.L1 && !ALTERNATIVE_GAS_TOKEN_ENABLED) {
-            uint256 l1MessengerBalance = address(L1_SCROLL_MESSENGER_PROXY_ADDR).balance;
+            address l1MessengerProxyAddr = contractsCfg.readAddress(".L1_SCROLL_MESSENGER_PROXY_ADDR");
```
Thegaram, line 275: I think this will break the step when we generate the addresses

scripts/deterministic/DeployScroll.s.sol
```diff
@@ -1033,7 +1041,7 @@ contract DeployScroll is DeterminsticDeployment {
         upgrade(L1_PROXY_ADMIN_ADDR, L1_ERC1155_GATEWAY_PROXY_ADDR, L1_ERC1155_GATEWAY_IMPLEMENTATION_ADDR);
     }
 
-    function deployL1GasTokenGateway() private gasToken(true) {
+    function deployL1GasTokenGateway() private gasToken(true) except(Layer.L2) {
```
Thegaram, line 1044: I don't really understand this, what will go wrong if we simulate this on L2?

yiweichi: Yes, that's because here:https://github.com/scroll-tech/scroll-contracts/blob/feat-deterministic-deployment/src/alternative-gas-token/L1GasTokenGateway.sol#L70

in the constructor of L1GasTokenGateway, it will call to get the decimal of gas token. On layer2, there are no gas tokne deployed.

Thegaram (approved):

PR #40, opened 2024-09-23
feat: gateway for l2 native erc20

zimpha:


colinlyguo (approved):

Thegaram (approved):

src/test/L1CustomERC20Gateway.t.sol
```diff
@@ -178,7 +178,16 @@ contract L1CustomERC20GatewayTest is L1GatewayTestBase {
         address recipient,
         bytes memory dataToCall
     ) public {
-        gateway.updateTokenMapping(address(l1Token), address(l2Token));
+        // message 0 is append here
+        gateway.updateTokenMapping{value: 1 ether}(address(l1Token), address(l2Token));
```
colinlyguo, line 182: why add this step?

zimpha: We only allow the update the mapping in L1 and relay the information L2 now. So the `updateTokenMapping` will result in a cross chain message, which need relay fee.

src/L2/gateways/L2CustomERC20Gateway.sol
```diff
@@ -109,11 +109,12 @@ contract L2CustomERC20Gateway is L2ERC20Gateway {
      ************************/
 
     /// @notice Update layer 2 to layer 1 token mapping.
+    ///
+    /// @dev To make the token mapping consistent with L1, this should be called from L1.
+    ///
     /// @param _l2Token The address of corresponding ERC20 token on layer 2.
     /// @param _l1Token The address of ERC20 token on layer 1.
-    function updateTokenMapping(address _l2Token, address _l1Token) external onlyOwner {
-        require(_l1Token != address(0), "token address cannot be 0");
```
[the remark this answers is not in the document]
zimpha: this function call only comes L1, trust L1 and don't check anything.

foundry.toml
```diff
@@ -7,9 +7,9 @@ libs = ["lib"]
 remappings = []                                               # a list of remappings
 libraries = []                                                # a list of deployed libraries to link against
 cache = true                                                  # whether to cache builds or not
-force = true                                                  # whether to ignore the cache (clean build)
-# evm_version = 'london'                                        # the evm version (by hardfork name)
-solc_version = '0.8.24'                                      # override for the solc version (setting this ignores `auto_detect_solc`)
+# force = true                                                # whether to ignore the cache (clean build)
```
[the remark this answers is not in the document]
zimpha: make compiling faster

src/test/L1CustomERC20Gateway.t.sol
colinlyguo: oh, I see. placed the comment in the wrong place.

the original problem is confused about the nonce change from 0 to 1, and it's because `updateTokenMapping` will send a cross-chain message (with nonce 0 in this test).

colinlyguo (approved):

PR #45, opened 2024-10-07
ci: add github action triggered manually to generate docker image

sbaizet-ledger:


.github/workflows/docker-release.yml
Thegaram, line 1: Is the idea that we merge this into `main` (because that's required for this trigger type) but only apply on `feat-deterministic-deployment`?

sbaizet-ledger: yes exactly.
I can add a branch filter

Thegaram: Got it. Not necessary, but let's add a comment saying that this workflow is used to build images from that branch.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,68 @@
...
+          REPOSITORY: scrolltech/scroll-stack-contracts
+        uses: docker/build-push-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: true
+          context: .
+          file: docker/Dockerfile.deploy
+          tags: |
+            ${{ env.REPOSITORY }}:deploy-${{ github.ref_name }}
+            ${{ env.REPOSITORY }}:latest
```
Thegaram, line 53: Let's remove `latest`. We don't need it, and the two images will overwrite each other.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,68 @@
...
+          REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          REPOSITORY: scrolltech/scroll-stack-contracts
+        uses: docker/build-push-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: true
+          context: .
+          file: docker/Dockerfile.deploy
+          tags: |
+            ${{ env.REPOSITORY }}:deploy-${{ github.ref_name }}
```
Thegaram, line 52: Can we use commit digest instead, as the tag? (Prefix of `github.sha`)

.github/workflows/docker-release.yml
sbaizet-ledger: not really, the same image will have 2 tags.
But okay to remove it

.github/workflows/docker-release.yml
sbaizet-ledger: not sure about this, as it is used on scroll-sdk, it will make it less clear which image is the newest. I find the semver tagging clearer.
Maybe we can get @dghelm opinion's on this

sbaizet-ledger: approved by daniel.

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Login to Dockerhub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build deploy image
+        id: build_deploy_image
+        env:
+          REGISTRY: ${{ steps.login-ecr.outputs.registry }}
```
Thegaram, line 44: Why do we have ECR here but Docker Hub above, is this correct?

.github/workflows/docker-release.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v3
+        with:
+          node-version: '21'
+
+      - name: Install dependencies
+        run: npm install
```
Thegaram, line 30: Not sure we need this, as I remember we install deps inside the container image. But if you tested and it works we can keep it.

Thegaram (dismissed):

.github/workflows/docker-release.yml
sbaizet-ledger: it's because we deploy 2 times, once to ECR and once to dockerhub.

Thegaram (approved):

yiweichi (approved):
LGTM

PR #58, opened 2024-10-29
feat: sgx verifier

zimpha:


src/L1/rollup/IScrollChain.sol
```diff
@@ -26,6 +26,22 @@ interface IScrollChain {
     /// @param withdrawRoot The merkle root on layer2 after this batch.
     event FinalizeBatch(uint256 indexed batchIndex, bytes32 indexed batchHash, bytes32 stateRoot, bytes32 withdrawRoot);
 
+    /// @notice Emitted when a batch is finalized by tee proof.
+    /// @param batchIndex The index of the batch.
+    event FinalizeBatchWithTEEProof(uint256 indexed batchIndex);
```
Thegaram, line 31: I'd rather keep the semantics of `FinalizeBatch` the same as batch finalization (currently it's still used for ZKP verification, which is not considered final anymore).

I.e. 3 events:
- `VerifyBatchWithZkp`
- `VerifyBatchWithTee`
- `FinalizeBatch`

(These last two will coincide -- but this will make sure that dapp/node logic that relies on `FinalizeBatch` will continue to work as expected.)

src/L1/rollup/IScrollChain.sol
```diff
@@ -140,4 +159,16 @@ interface IScrollChain {
         bytes32 withdrawRoot,
         bytes calldata aggrProof
     ) external;
+
+    /// @notice Finalize a list of committed batches (i.e. bundle) on layer 1 with TEE proof.
+    /// @param batchHeader The header of last batch in current bundle, see the encoding in comments of `commitBatch.
```
Thegaram, line 164:
```suggestion
    /// @param batchHeader The header of last batch in current bundle, see the encoding in comments of `commitBatch`.
```

src/L1/rollup/ScrollChain.sol
```diff
@@ -583,10 +623,83 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        // check withdraw root and state root match with zk proof
+        if (withdrawRoots[_batchIndex] != _withdrawRoot || cachedStateRoot != _postStateRoot) {
+            unresolvedState.batchIndex = _batchIndex;
+            unresolvedState.stateRoot = _postStateRoot;
+            unresolvedState.withdrawRoot = _withdrawRoot;
+            emit StateMismatch(_batchIndex, _postStateRoot, _withdrawRoot);
+            return;
+        }
+
+        lastTeeFinalizedBatchIndex = _batchIndex;
```
Thegaram, line 669: Again, I would prefer `**VerifiedBatchIndex`, since either of the two proofs is not considered "finalized" anymore. So either `lastTeeVerifiedBatchIndex`, or `lastFinalizedBatchIndex` (since finalized == zkp + tee) would be better here.

src/L1/rollup/ScrollChain.sol
Thegaram, line 1: We should use `lastTeeFinalizedBatchIndex` in `isBatchFinalized` and `revertBatch`.

src/L1/rollup/ScrollChain.sol
```diff
@@ -583,10 +623,83 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+    function finalizeBundleWithTeeProof(
+        bytes calldata _batchHeader,
+        bytes32 _postStateRoot,
+        bytes32 _withdrawRoot,
+        bytes calldata _teeProof
+    ) external override OnlyProver whenFinalizeNotPaused {
+        // compute pending batch hash and verify
+        (uint256 batchPtr, bytes32 _batchHash, uint256 _batchIndex, ) = _loadBatchHeader(_batchHeader);
+
+        // check if this batch is finalized by zk proof.
```
Thegaram, line 636:
```suggestion
        // check if this state root is verified by zk proof.
```

Just a nitpick. It's still possible that the batch is "finalized" by zkp but it's in the middle of the bundle, so we do not have it stored in state.

src/L1/rollup/ScrollChain.sol
Thegaram, line 1: Do we want to make it possible (through governance vote) to turn off the TEE verification layer?

src/L1/rollup/ScrollChain.sol
```diff
@@ -583,10 +623,83 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+    ///
+    /// @param useSGXState Whether we want to use the state root from SGX prover.
+    function resolveStateMismatch(bool useSGXState) external onlyOwner {
+        UnresolvedState memory state = unresolvedState;
+        if (state.batchIndex == 0) {
+            revert ErrorNoUnresolvedState();
+        }
+        if (useSGXState) {
+            finalizedStateRoots[state.batchIndex] = state.stateRoot;
+            withdrawRoots[state.batchIndex] = state.withdrawRoot;
```
Thegaram, line 690: It seems to me that in this case we must reset `lastZkpVerifiedBatchIndex` (currently called `lastFinalizedBatchIndex`), since subsequent verified batches are invalidated.

src/sgx-verifier/AttestationVerifier.sol
```diff
@@ -0,0 +1,207 @@
...
+
+    /// @notice The address of automata's DCAP Attestation contract.
+    IDcapAttestation public immutable attestationVerifier;
+
+    /*********************
+     * Storage Variables *
+     *********************/
+
+    /// @dev The list of trusted enclaves.
+    EnumerableSet.Bytes32Set private trustedUserMrEnclave;
```
Thegaram, line 46: Is this basically the "program hash"? Or the identity of the (Intel) signer?

src/sgx-verifier/AttestationVerifier.sol
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /************************
+     * Restricted Functions *
+     ************************/
+
+    /// @notice Update the status of a signer.
+    /// @param _mrSigner The signer to update.
+    /// @param _status The new status to update. If it is `true`, the signer is trusted.
+    function updateMrSigner(bytes32 _mrSigner, bool _status) external onlyOwner {
```
Thegaram, line 118: What is exactly `mrSigner`? How and when will we call this?

src/sgx-verifier/AttestationVerifier.sol
```diff
@@ -0,0 +1,207 @@
...
+            trustedUserMrSigner.remove(_mrSigner);
+        }
+
+        emit UpdateMrSigner(_mrSigner, _status);
+    }
+
+    /// @notice Update the status of an enclave.
+    /// @param _mrEnclave The enclave to update.
+    /// @param _status The new status to update. If it is `true`, the enclave is trusted.
+    function updateMrEnclave(bytes32 _mrEnclave, bool _status) external onlyOwner {
```
Thegaram, line 132: Same question for MrEnclave.

src/sgx-verifier/SGXVerifier.sol
```diff
@@ -0,0 +1,220 @@
...
+            numBatches,
+            prevStateRoot,
+            prevBatchHash,
+            postStateRoot,
+            batchHash,
+            postWithdrawRoot
+        );
+        address prover = ECDSA.recover(hash, bundleProof);
+
+        if (attestedProvers[prover].validUntil < block.timestamp) {
```
Thegaram, line 164: So in this version, any registered prover can submit an SGX proof, right?

src/sgx-verifier/SGXVerifier.sol
Thegaram, line 1: How will we deal with protocol upgrades (multiple batch versions)?

Will we deploy multiple instances of `SGXVerifier` with different `attestationVerifier` contracts? Or will they reuse the same `attestationVerifier`?

src/L1/rollup/ScrollChain.sol
zimpha: It would make the code more complex if we add such flag in the contract. But let @icemelon decides whether we should do it or not.

src/sgx-verifier/AttestationVerifier.sol
zimpha: `MrEnclave` is the program hash.

src/sgx-verifier/AttestationVerifier.sol
zimpha: `mrSigner` is the one who compiles the enclave problem and sign the program hash.

src/sgx-verifier/AttestationVerifier.sol
zimpha: answered in https://github.com/scroll-tech/scroll-contracts/pull/58#discussion_r1828794676

src/sgx-verifier/SGXVerifier.sol
zimpha: Yeah, we allow any registered prover submit an SGX proof in current version. This will make the contracts code clean.

src/sgx-verifier/SGXVerifier.sol
zimpha: We will have a `MultipleVersionRollupVerifier` contract deployed only for `SGXVerifier` and the version managed in this contract. So each time we upgrade batch version, both `MultipleVersionRollupVerifier` for zkp and tee should be updated.

For different `SGXVerifier`, I believe we can share the same `attestationVerifier`.

src/sgx-verifier/AttestationVerifier.sol
Thegaram: Got it. Since we already sign the transaction when we update `mrEnclave` (the program hash), what is the use of `mrSigner`?

My understanding was this signing entity would allow us to update the program hash just by verifying the signature. Useful when the program changes frequently. But it seems we don't need this functionality.

src/L1/rollup/ScrollChain.sol
zimpha: agreed, code added

src/sgx-verifier/AttestationVerifier.sol
zimpha: `mrSigner` and `mrEnclave` is part of the attestation report, we can recover them from the report and check whether the recovered one is whitelisted. Just some cross validation to make sure everything is ok.

src/sgx-verifier/SGXVerifier.sol
Thegaram: > For different SGXVerifier, I believe we can share the same attestationVerifier.

I'm not sure. Basically we'd need `batchVersion ==> mrEnclave` mapping implemented by `MultipleVersionRollupVerifier`. (In practice it's `batchVersion ==> sgxVerifier ==> attestationVerifier`.)

src/L1/rollup/ScrollChain.sol
```diff
@@ -569,24 +621,121 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        }
+        (, , uint256 _batchIndex, uint256 _totalL1MessagesPoppedOverall) = _loadBatchHeader(_batchHeader);
+        if (_batchIndex != state.batchIndex) revert ErrorBatchIndexMismatch();
+        if (useSGXState) {
+            finalizedStateRoots[state.batchIndex] = state.stateRoot;
+            withdrawRoots[state.batchIndex] = state.withdrawRoot;
+            emit ResolveState(state.batchIndex, state.stateRoot, state.withdrawRoot);
+
+            // reset zkp verified batch index, zk prover need to reprove everything after this batch
+            lastZkpVerifiedBatchIndex = state.batchIndex;
```
Thegaram, line 721: @colinlyguo any thoughts what we'd need to add to be able to recover from this relatively painlessly? Maybe just listen to `ResolveState` events in `rollup-relayer` and update db accordingly? (Though in this case most likely we'll need to fix either zk or sgx prover.)

src/mocks/ScrollChainMockFinalize.sol
Thegaram, line 1: So with `MockFinalize` mode we'd still need to run an SGX prover?

Thegaram (dismissed):

src/mocks/ScrollChainMockFinalize.sol
zimpha: Yeah, running a SGX prover is cheap. I think it is ok to run it in sepolia.

src/L1/rollup/ScrollChain.sol
```diff
@@ -569,24 +621,121 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
         // our off-chain service will make sure all unfinalized batches have the same batch version.
-        IRollupVerifier(verifier).verifyBundleProof(batchVersion, _batchIndex, _aggrProof, _publicInput);
+        IRollupVerifier(zkpVerifier).verifyBundleProof(batchVersion, _batchIndex, _aggrProof, _publicInput);
 
         // store in state
         // @note we do not store intermediate finalized roots
-        lastFinalizedBatchIndex = _batchIndex;
+        lastZkpVerifiedBatchIndex = _batchIndex;
         finalizedStateRoots[_batchIndex] = _postStateRoot;
         withdrawRoots[_batchIndex] = _withdrawRoot;
```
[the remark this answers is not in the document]
zimpha: We can override the function of `finalizedStateRoots` and `withdrawRoots`, to make it return `bytes(0)` for all indices after `lastFinalizedBatchIndex`.

src/L1/rollup/IScrollChain.sol
```diff
@@ -19,13 +19,44 @@ interface IScrollChain {
...
+        bytes32 stateRoot,
+        bytes32 withdrawRoot
+    );
+
+    /// @notice Emitted when a batch is verified by tee proof
+    /// @dev Tee proof always comes after zk proof. If they match, the state root and withdraw root are the same.
+    ///      If they mismatch, we will emit `StateMismatch` instead. Therefore, the `stateRoot` and `withdrawRoot`
+    ///      is not included in this event.
+    /// @param batchIndex The index of the batch.
+    event VerifyBatchWithTee(uint256 indexed batchIndex);
```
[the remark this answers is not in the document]
zimpha: for saving some gas and we can infer all other arguments based on other events.

src/sgx-verifier/SGXVerifier.sol
zimpha: > (In practice it's `batchVersion` ==> `sgxVerifier` ==> `attestationVerifier`.)

true, we can decide it later. It is only related to deployment.

src/mocks/ScrollChainMockFinalize.sol
Thegaram: Since the verifier is set as `address(0)`, does that mean that we just need someone to call `finalizeBundleWithTeeProof` (e.g. in `rollup-relayer`, when fake finalize mode is enabled), and the `verifyBundleProof` call will be a no-op?

zimpha: seems that I forget to change the constructor. The `verifier` should not be set as `address(0)`. I did change `ScrollChainMockBlob` since it is used in unit tests. I'm thinking we should also add unit tests for `ScrollChainMockFinalize`.

Thegaram: I see two options two make `ScrollChainMockFinalize` useful:
1. Use `address(0)` as verifier so that we bypass the whole verification step. Then, `rollup-relayer` can call `finalizeBundleWithTeeProof` when fake finalization is enabled.
2. Keep the verifier, but somehow bypass `verifyAttestation`, so that we can run the "SGX" prover on normal hardware. (Setting up SGX hardware is too complex for a devnet.)

zimpha: For option 2, we need test the feature in sepolia first before goes to mainnet. After test we can do anything we like, like bypass `verifyAttestation` or fake finalization on tee proof.

src/L1/rollup/ScrollChain.sol
```diff
@@ -218,13 +260,24 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         emit UpdateMaxNumTxInChunk(0, _maxNumTxInChunk);
     }
 
+    function initializeV2() external reinitializer(2) {
+        uint256 cachedIndex = lastZkpVerifiedBatchIndex;
+        lastTeeVerifiedBatchIndex = cachedIndex;
+        emit VerifyBatchWithTee(cachedIndex);
```
colinlyguo, line 266: Would another event name to differentiate from real tee verify a better choice?

src/L1/rollup/ScrollChain.sol
```diff
@@ -569,24 +621,121 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+            return;
+        }
+
+        lastTeeVerifiedBatchIndex = _batchIndex;
+
         // Pop finalized and non-skipped message from L1MessageQueue.
         _finalizePoppedL1Messages(_totalL1MessagesPoppedOverall);
 
+        emit VerifyBatchWithTee(_batchIndex);
         emit FinalizeBatch(_batchIndex, _batchHash, _postStateRoot, _withdrawRoot);
```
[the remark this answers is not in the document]
colinlyguo: chain-monitor (does not monitor batch finalize events, only monitors finalize deposit/withdraw events), bridge-history (withdraw proof), and l2geth-verifier (finalize batch verification) are compatible. I believe sync-from-DA feature is also compatible.

We can keep a close eye on monitoring these components after the upgrade as well.

src/L1/rollup/ScrollChain.sol
colinlyguo: > Maybe just listen to ResolveState events in rollup-relayer and update db accordingly

src/L1/rollup/ScrollChain.sol
```diff
@@ -218,13 +319,75 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        lastTeeVerifiedBatchIndex = cachedIndex;
+        emit VerifyBatchWithTee(
+            cachedIndex,
+            committedBatches[cachedIndex],
+            finalizedStateRoots[cachedIndex],
+            withdrawRoots[cachedIndex]
+        );
+
+        // initialize the first element in the array
+        bundleSize.push(BundleSizeStruct(_bundleSize, uint128(cachedIndex)));
```
Thegaram, line 334: How will the upgrade happen? Basically right after executing the upgrade we'll need to reset all pending bundles in the rollup-relayer, right?

src/L1/rollup/ScrollChain.sol
```diff
@@ -157,6 +220,31 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+    UnresolvedState public unresolvedState;
+
+    /// @notice The state for bundle size.
+    /// @dev This is enabled after Euclid upgrade.
+    /// @dev Assume the list is `[(s[1], b[1]), (s[2], b[2]), ..., (s[n], b[n])]`,  where `s[i]` is the bundle size
+    ///      and `b[i]` is the start batch index. Then for each `i > 1`, we should have `b[i] > b[i - 1]` and
+    ///      `(b[i] - b[i - 1]) % s[i - 1] = 0`.
+    ///      If a bundle has end batch range `[x, y]`, we need to find last `i` such that `y > b[i]`. And the following
+    ///      should be satisfied: `(y - b[i]) % s[i] = 0` and `y - x + 1 = s[i]`.
+    BundleSizeStruct[] public bundleSize;
```
colinlyguo, line 246: does this mean rollup-relayer tee-prover would read the slots to decide bundle size? since currently tee-prover is not only an event-listening client.

src/L1/rollup/ScrollChain.sol
```diff
@@ -645,10 +883,46 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         }
     }
 
+    /// @notice Update the bundle size
+    /// @param size The new bundle size.
+    /// @param batchIndex The start batch index for new bundle size.
+    function updateBundleSize(uint128 size, uint128 batchIndex) external onlyOwner {
```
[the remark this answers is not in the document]
zimpha: ok, make the last item updatable.

src/L1/rollup/ScrollChain.sol
zimpha: for rollup-relayer, local configuration is fine, since only us run it. Tee host can read from contract.

src/L1/rollup/ScrollChain.sol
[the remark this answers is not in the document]
zimpha: I choose to add comments about this. Overriding those two functions make the code more complex.

src/L1/rollup/ScrollChain.sol
```diff
@@ -218,13 +319,75 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+                block.timestamp >= batchCommittedTimestamp[_batchIndex] + proofDelayTimestamp;
+        } else {
+            return _batchIndex <= lastFinalizedBatchIndex();
+        }
+    }
+
+    /// @dev Get bundle size with given end batch index.
+    /// @param batchIndex The end batch index of given bundle.
+    /// @return size The size of the given bundle.
+    function getBundleSizeGivenEndBatchIndex(uint256 batchIndex) public view returns (uint256) {
```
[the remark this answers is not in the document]
zimpha: In contract, the given `batchIndex` is always valid one. Extra checking will cost more gas.

src/L1/rollup/ScrollChain.sol
zimpha: Currently, the bundle size is about 100, right? We can use this for the rollup-relayer, and no need to do anything switch for rollup-relayer during upgrade.

src/L1/rollup/ScrollChain.sol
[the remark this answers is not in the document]
colinlyguo: oh got it.

src/L1/rollup/ScrollChain.sol
colinlyguo: added relevant tweaks of bundle-proposer in this PR: https://github.com/scroll-tech/scroll/pull/1566

PR #68, opened 2024-12-05
feat: enforced batch mode

zimpha:


src/libraries/codec/BatchHeaderV3Codec.sol
```diff
@@ -46,6 +46,15 @@ library BatchHeaderV3Codec {
         }
     }
 
+    /// @notice Get the blob versioned hash of the batch header.
```
jonastheis, line 49: wrong comment

jonastheis (approved):
LGTM!

with the new batch design this will need to change again but the main mechanism of
1. track time of latest finalized batch
2. if too much time elapsed allow anyone to commit batch + proof and disallow sequencer to submit via usual methods
3. only allow security council to deactivate this enforced/permissionless mode

should stay the same and just the interface of what/how a batch needs to be passed should change.

src/L1/rollup/ScrollChain.sol
```diff
@@ -172,6 +197,11 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         _;
     }
 
+    modifier whenEnforcedBatchNotEnable() {
```
Thegaram, line 200:
```suggestion
    modifier whenEnforcedBatchNotEnabled() {
```

src/L1/rollup/ScrollChain.sol
```diff
@@ -218,6 +248,28 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+            step >>= 1;
+            while (step > 0) {
+                if (committedBatches[index + step] != bytes32(0)) {
+                    index += step;
+                }
+                step >>= 1;
+            }
+        }
+        // initialize enforcedBatchParameters
+        enforcedBatchParameters.maxDelayEnterEnforcedMode = _maxDelayEnterEnforcedMode;
```
Thegaram, line 268: Any way to update this?

src/L1/rollup/ScrollChain.sol
```diff
@@ -126,6 +135,18 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
     /// @notice The address of RollupVerifier.
     address public immutable verifier;
 
+    /***********
+     * Structs *
+     ***********/
+
+    struct EnforcedBatchParameters {
```
Thegaram, line 142: Would this add +1 SSTORE cost for normal commit batch and finalize bundle txs?

src/L1/rollup/ScrollChain.sol
```diff
@@ -436,151 +363,61 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        external
+    {
+        // we use BatchHeaderV3Codec for finalizeStruct.batchHeader
+        EnforcedBatchParameters memory parameters = enforcedBatchParameters;
+        if (!parameters.enforcedModeEnabled) {
+            if (parameters.lastFinalizeTimestamp + parameters.maxDelayEnterEnforcedMode < block.timestamp) {
+                // explicit set enforce batch enable
+                parameters.enforcedModeEnabled = true;
+                // reset `lastCommittedBatchIndex`
+                parameters.lastCommittedBatchIndex = uint64(lastFinalizedBatchIndex);
```
Thegaram, line 394: I think we've discussed this before but I forgot the conclusion; doesn't this make "fast L2 finality" impossible? @jonastheis

In the future with the infallible chain derivation, we could just have a permissionless fallback for batch commitment and finalization (but not need to revert unfinalized batches).

src/L1/rollup/ScrollChain.sol
Thegaram, line 1: Overall looks good, but this will have a lot of conflicts with other changes planned for this upgrade. Do you want to still submit them as separate PRs and then merge them into a single unified PR that we share with the auditors?

zimpha: yes, I will create a branch and all related PR will merged into this branch.

src/L1/rollup/ScrollChain.sol
zimpha: my guess is about `4600` gas.

src/L1/rollup/ScrollChain.sol
zimpha: I assume it won't be updated when we set it. Sure I can add a method to update it.

Thegaram: @jonastheis do you think we might want to update this in the future? Seems safer to me to keep this updatable, at least for the 1st version.

jonastheis: Yeah I think would be better if we (I assume Security council) can update it

src/L1/rollup/ScrollChain.sol
jonastheis: With chain derivation we will need to change the mechanism anyway. So I think it is okay to revert here.

PR #72, opened 2025-01-15
feat: add system config consensus to deprecate clique

ranchalp:
In conjunction with [go-ethereum#1102](https://github.com/scroll-tech/go-ethereum/pull/1102).


src/L1/system-contract/SystemSignerRegistry.sol
```diff
@@ -0,0 +1,65 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
+ * @dev Example "SystemSignerRegistry" storing `(startBlock, signer)` pairs.
+ *      The getSigners() function returns parallel arrays for block numbers and addresses,
+ *      
+ */
+contract SystemSignerRegistry is Ownable {
+    struct Signer {
+        uint64 startBlock;
```
Thegaram, line 13: Didn't we decide against having the `startBlock` here?

src/L1/system-contract/SystemSignerRegistry.sol
```diff
@@ -0,0 +1,65 @@
...
+        address signer;
+    }
+
+    Signer[] private signers;
```
Thegaram, line 17:
```suggestion
    mapping (address => bool) public isAuthorizedSigner;
```

src/L1/system-contract/SystemSignerRegistry.sol
```diff
@@ -0,0 +1,65 @@
...
+pragma solidity =0.8.24;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
+ * @dev Example "SystemSignerRegistry" storing `(startBlock, signer)` pairs.
+ *      The getSigners() function returns parallel arrays for block numbers and addresses,
+ *      
+ */
+contract SystemSignerRegistry is Ownable {
```
Thegaram, line 11: Do we want to make this upgradable?

Thegaram, line 11: Why not model this after OP's? https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/SystemConfig.sol (see for example `_setUnsafeBlockSigner`).

@zimpha what do you think about storing multiple configs here (including gas coefficients, max gas limit, etc.)? Or is that too expensive gas-wise?

src/L1/system-contract/SystemSignerRegistry.sol
Thegaram: Need to consider how `l2geth` will read this contract.
- If it just polls the latest state asking "am I authorized?", then a mapping is suitable.
- If it just polls the latest state asking "what is the list of current authorized signers?", then an array is suitable.
- If it fetches events, then I'd still go with mapping.

src/L1/system-contract/SystemSignerRegistry.sol
zimpha: we can add these config here, we can pack related parameters into single `uint256`. Then the gas overhead will be minimum.

src/L1/system-contract/SystemSignerRegistry.sol
zimpha: will we have multiple authorized signer in the same time?

src/L1/system-contract/SystemSignerRegistry.sol
jonastheis: iirc we discussed it [here](https://scrollco.slack.com/archives/C02PKAQ2430/p1735815568292809?thread_ts=1735809529.760869&cid=C02PKAQ2430) but didn't reach a conclusion

src/L1/system-contract/SystemSignerRegistry.sol
ranchalp: No, we should not have multiple authorized signers at the same time.

Thegaram: Having multiple makes key rotation a bit more convenient. But if we want to keep it simple, we can also just have one (`currentAuthorizedSigner`).

src/L1/system-contract/SystemSignerRegistry.sol
Thegaram: I meant the gas overhead of `CALL` (e.g. `systemContract.getGasParams()`), not just `SLOAD`, is that acceptable?

src/L1/system-contract/SystemSignerRegistry.sol
Thegaram: >  The reason is that otherwise the zkproof must accept any of two valid keys

Which zkproof do you mean? If you mean when we finalize bundles, currently we don't prove anything about the block signatures.

ranchalp: Ah, I see. Let me put it back with one signer then and forget about this orderly switching with blockNum for now (bringing it back in the future probably).

ranchalp:
Added back one signer only

zimpha:
refactor to https://github.com/scroll-tech/scroll-contracts/pull/75

PR #74, opened 2025-02-13
feat: euclid phase-2 contract changes

zimpha:
This PR contains all changes for the planned `EuclidV2` upgrade. Specifically,
- [Reduce commit cost](https://www.notion.so/scrollzkp/Reduce-Commit-Cost-1137792d22af808aa5b9c02150cdff4d?pvs=4): we reduce the cost to commit batches significantly by
  1. move calldata (chunks and block contexts) to blob -> no notion of chunks anymore 
  2. chaining of L1 messages with `MessageQueueV2`: instead of iterating through L1 messages and computing the hash at commit time, L1 messages are chained with a rolling hash. so that the last hash represents a commitment to the entire queue.
  3. move point evaluation to circuit
  4. commit multiple batches in a single transaction
- [Enforced transactions](https://www.notion.so/scrollzkp/Enforced-Transactions-d6a0d9797bb74b16a84d4a1b35c73e4b?pvs=4): transactions are appended to `MessageQueueV2` together with the time they have been submitted to L1. If not included/finalized for `maxDelayMessageQueue`, the operator can not finalize any more batches until including those messages. If it continues to censor messages permissionless mode will eventually activate.
- [Enforced/permissionless batches](https://www.notion.so/scrollzkp/Enforced-Batches-7e5c1ac7264d4c6aaac1a45fbf6dce4f?pvs=4): if no batch has been finalized for `maxDelayEnterEnforcedMode`, batch submission becomes permissionless. Anyone can submit a batch together with a proof.
- [Deprecate Clique](https://www.notion.so/scrollzkp/Deprecate-Clique-1157792d22af80e08837ea02010a778e?pvs=4): we introduce a `SystemConfig` contract that contains the address of the currently valid signer for L2 blocks. L2 nodes can read and will only accept L2 blocks signed by the given address.


jonastheis (commented):
Some questions:
- maybe I missed it but do we ensure that all messages on `MessageQueueV1` are finalized before we allow finalization of `MessageQueueV2` messages?
- for enforced transactions: don't we need to check that all messages on the queue that are older than a certain threshold are included when finalizing? otherwise we don't actually force the sequencer to include them. it could keep censoring forever and permissionless mode would not activate either.

src/L1/rollup/ScrollChain.sol
```diff
@@ -650,6 +771,74 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         if (numTransactions != 0) revert ErrorV5BatchContainsTransactions();
     }
 
+    function _commitBatchesPostEuclid(uint8 version, bool onlyOne) internal returns (bytes32) {
+        if (version < 7) {
+            // only accept version >= 7
+            revert ErrorIncorrectBatchVersion();
+        }
+        uint256 lastCommittedBatchIndex = miscData.lastCommittedBatchIndex;
+        bytes32 parentBatchHash = committedBatches[lastCommittedBatchIndex];
```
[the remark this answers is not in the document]
kadima233: yeah, it would be a problem. I can add variable `batchIndex` to the function.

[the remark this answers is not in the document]
zimpha: yeah, it would be a problem. I can add a variable `parentBatchIndex` to the parameters.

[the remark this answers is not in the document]
jonastheis: I think passing the hash of the last batch we expect to be created is safer

src/mocks/ScrollChainMockFinalize.sol
```diff
@@ -37,6 +33,19 @@ contract ScrollChainMockFinalize is ScrollChain {
         );
 
         // actions after verification
-        _afterFinalizeBatch(batchIndex, batchHash, totalL1MessagesPoppedOverall, postStateRoot, withdrawRoot);
+        _afterFinalizeBatch(batchIndex, batchHash, totalL1MessagesPoppedOverall, postStateRoot, withdrawRoot, true);
```
Thegaram, line 36: Why do we use V1 queue and old public input format for enforced batches? The old prover cannot prove the new evm

src/L1/rollup/ScrollChain.sol
```diff
@@ -408,28 +467,62 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+            (uint256 maxDelayEnterEnforcedMode, uint256 maxDelayMessageQueue) = SystemConfig(systemConfig)
+                .enforcedBatchParameters();
+            uint256 lastUnfinalizedMessageTime = IL1MessageQueueV2(messageQueueV2)
+                .getFirstUnfinalizedMessageEnqueueTime();
+            if (
+                lastUnfinalizedMessageTime + maxDelayMessageQueue < block.timestamp ||
+                cachedMiscData.lastFinalizeTimestamp + maxDelayEnterEnforcedMode < block.timestamp
+            ) {
+                // explicit set enforce batch enable
+                cachedMiscData.enforcedModeEnabled = true;
```
Thegaram, line 503: Emit event here and in `disableEnforcedBatch`.

src/L1/rollup/ScrollChain.sol
```diff
@@ -408,28 +467,62 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+    }
+
+    /// @inheritdoc IScrollChain
+    /// @dev We only consider batch version >= 7 here.
+    function commitAndFinalizeBatch(
+        uint8 version,
+        bytes32 parentBatchHash,
+        FinalizeStruct calldata finalizeStruct
+    ) external {
+        // we use BatchHeaderV3Codec for finalizeStruct.batchHeader
```
Thegaram, line 491: BatchHeaderV7Codec?

src/L1/gateways/EnforcedTxGateway.sol
Thegaram, line 1: Any reason to keep the EOA check? Can we remove it?

src/L1/L1ScrollMessenger.sol
```diff
@@ -343,19 +356,19 @@ contract L1ScrollMessenger is ScrollMessengerBase, IL1ScrollMessenger {
...
+        uint256 _fee = IL1MessageQueueV2(messageQueueV2).estimateCrossDomainMessageFee(_gasLimit);
         require(msg.value >= _fee + _value, "Insufficient msg.value");
         if (_fee > 0) {
             (bool _success, ) = feeVault.call{value: _fee}("");
             require(_success, "Failed to deduct the fee");
         }
 
         // append message to L1MessageQueue
-        IL1MessageQueue(messageQueue).appendCrossDomainMessage(counterpart, _gasLimit, _xDomainCalldata);
+        IL1MessageQueueV2(messageQueueV2).appendCrossDomainMessage(counterpart, _gasLimit, _xDomainCalldata);
```
Thegaram, line 371: Just to confirm: We will need to (1) initialize `messageQueueV2`, (2) upgrade `L1ScrollMessenger`, (3) upgrade `EnforcedTxGateway` in a single atomic transaction, or things might break, is that correct?

src/batch-bridge/L1BatchBridgeGateway.sol
Thegaram, line 1: Is this still used?

src/L1/gateways/EnforcedTxGateway.sol
zimpha: we can remove it.

src/L1/L1ScrollMessenger.sol
zimpha: yeah, it should be a single atomic transaction.

src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: not used, but we can keep the contracts in the repo.

src/L1/gateways/EnforcedTxGateway.sol
```diff
@@ -91,10 +93,13 @@ contract EnforcedTxGateway is OwnableUpgradeable, ReentrancyGuardUpgradeable, Pa
         uint256 _gasLimit,
         bytes calldata _data
     ) external payable whenNotPaused {
-        // solhint-disable-next-line avoid-tx-origin
-        require(msg.sender == tx.origin, "Only EOA senders are allowed to send enforced transaction");
+        address sender = _msgSender();
+        if (sender != tx.origin) {
+            // alias smart contract account, consistent with message from L1ScrollMessenger.
+            sender = AddressAliasHelper.applyL1ToL2Alias(sender);
```
Thegaram, line 99: I wonder about the implications here. Does this mean that smart contract accounts have no guaranteed way to force exit?
- For 7702 delegated accounts, they can force exit by using the original private key to send the message.
- For other smart accounts, they should authorize the aliased address on L2 in advance.

zimpha: usually, the smart contract accounts are controlled by some EOA. The EOA can call functions in the smart contract accounts to force exit.

src/L1/system-contract/SystemConfig.sol
```diff
@@ -0,0 +1,99 @@
...
+     ***********/
+
+    /// @notice Parameters for message queue. The compiler will pack this struct into single `bytes32`.
+    /// @param maxGasLimit The maximum gas limit allowed for each L1 message.
+    /// @param baseFeeOverhead The overhead used to calculate l2 base fee.
+    /// @param baseFeeScalar The scalar used to calculate l2 base fee.
+    struct MessageQueueParameters {
+        uint32 maxGasLimit;
+        uint112 baseFeeOverhead;
+        uint112 baseFeeScalar;
```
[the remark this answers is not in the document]
zimpha: just call `updateMessageQueueParameters` function to update the parameters. If you mean the value, it should be the same as current l2geth.

src/L1/rollup/L1MessageQueueV2.sol
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    /// @inheritdoc IL1MessageQueueV2
+    function getMessageRollingHash(uint256 queueIndex) external view returns (bytes32 hash) {
+        (hash, ) = _loadAndDecodeRollingHash(queueIndex);
+    }
+
+    /// @inheritdoc IL1MessageQueueV2
+    function estimatedL2BaseFee() public view returns (uint256) {
+        (, uint256 overhead, uint256 scalar) = SystemConfig(systemConfig).messageQueueParameters();
```
Thegaram, line 169:
```suggestion
        (, uint112 overhead, uint112 scalar) = SystemConfig(systemConfig).messageQueueParameters();
```

Easier to see that `unchecked` is safe if we keep the original types here

src/L1/system-contract/SystemConfig.sol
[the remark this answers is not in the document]
Thegaram: Yeah, I meant the latter. So when we update these, we need to update both in l2geth and in the L1 SystemConfig contract, is that right?

src/L1/rollup/L1MessageQueueV2.sol
zimpha: you mean we assume `estimatedL2BaseFee` won't exceed `uint112`? But in `return (block.basefee * scalar) / PRECISION + overhead;`, all types will be converted to `uint256` first. So explicitly use `uint256` would save some gas.

I can add  more comments here to explain the reason why `unchecked` is safe.

src/L1/system-contract/SystemConfig.sol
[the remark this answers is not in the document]
zimpha: yes, both l2geth and  the SystemConfig contract. And it is ok to do it asynchronously, small delay is fine.

src/L1/rollup/ScrollChain.sol
```diff
@@ -478,41 +473,53 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        // we use BatchHeaderV7Codec for finalizeStruct.batchHeader
         ScrollChainMiscData memory cachedMiscData = miscData;
         if (!cachedMiscData.enforcedModeEnabled) {
-            if (cachedMiscData.lastFinalizeTimestamp + cachedMiscData.maxDelayEnterEnforcedMode < block.timestamp) {
+            (uint256 maxDelayEnterEnforcedMode, uint256 maxDelayMessageQueue) = SystemConfig(systemConfig)
+                .enforcedBatchParameters();
+            uint256 lastUnfinalizedMessageTime = IL1MessageQueueV2(messageQueueV2)
+                .getFirstUnfinalizedMessageEnqueueTime();
+            if (
+                lastUnfinalizedMessageTime + maxDelayMessageQueue < block.timestamp ||
```
jonastheis, line 501: Just do double check: is this correct? we enable permissionless mode in 2 cases:
1. message is too long on the queue, OR
2. no batch finalized for too long

These two are independent of each other.

src/L1/rollup/ScrollChain.sol
```diff
@@ -771,16 +775,30 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        bool onlyOne
+    ) internal returns (bytes32) {
         if (version < 7) {
             // only accept version >= 7
             revert ErrorIncorrectBatchVersion();
         }
+
         uint256 lastCommittedBatchIndex = miscData.lastCommittedBatchIndex;
-        bytes32 parentBatchHash = committedBatches[lastCommittedBatchIndex];
+        if (parentBatchHash != committedBatches[lastCommittedBatchIndex]) {
```
jonastheis, line 793: What I meant with my [comment](https://github.com/scroll-tech/scroll-contracts/pull/74#discussion_r1955624085) was to take the last batch we expect to be generated in this commit transaction (ie the batch hash we're writing at the end in `committedBatches[lastCommittedBatchIndex] = parentBatchHash;`). This way we know for sure that all batches we committed are the same or revert.

src/L1/rollup/ScrollChain.sol
```diff
@@ -804,13 +822,25 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+    function _finalizeBundlePostEuclidV2(
         bytes calldata batchHeader,
         uint256 lastProcessedQueueIndex,
         bytes32 postStateRoot,
         bytes32 withdrawRoot,
         bytes calldata aggrProof
     ) internal {
+        if (
+            IL1MessageQueueV1(messageQueueV1).nextUnfinalizedQueueIndex() !=
+            IL1MessageQueueV2(messageQueueV2).firstCrossDomainMessageIndex()
```
jonastheis, line 840: Not sure how expensive that is but we need to do it every time we finalize. might be worth storing a bool that gets set after we finalized the first V7 batch? then we only need to check that going forward

src/L1/rollup/ScrollChain.sol
zimpha: I think they are the same. If `parentBatchHash` is fixed, all subsequent batch hashes are determined.

src/L1/rollup/ScrollChain.sol
zimpha: the first condition `lastUnfinalizedMessageTime + maxDelayMessageQueue < block.timestamp` check "message is too long on the queue"
and the second condition `cachedMiscData.lastFinalizeTimestamp + maxDelayEnterEnforcedMode < block.timestamp` check "no batch finalized for too long".

two conditions are OR relationship.

src/L1/rollup/ScrollChain.sol
zimpha: Yes, add a additional flag would save some gas.

src/L1/rollup/ScrollChain.sol
Thegaram: I think Jonas's suggestion is something like this.

```diff
- function commitBatchesPostEuclidV2(uint8 version, bytes32 parentBatchHash)
+ function commitBatchesPostEuclidV2(uint8 version, bytes32 parentBatchHash, bytes32 lastBatchHash)
// ...
- committedBatches[lastCommittedBatchIndex] = parentBatchHash;
+ if (parentBatchHash != lastBatchHash) revert InconsistentBatchHash(parentBatchHash, lastBatchHash);
+ committedBatches[lastCommittedBatchIndex] = parentBatchHash;
```

zimpha: yeah, I know what he means.
But if you have `parentBatchHash` and list of blob hashes, `lastBatchHash` is fixed. No need to provide another parameter `lastBatchHash`.

src/L1/gateways/EnforcedTxGateway.sol
```diff
@@ -91,10 +93,13 @@ contract EnforcedTxGateway is OwnableUpgradeable, ReentrancyGuardUpgradeable, Pa
         uint256 _gasLimit,
         bytes calldata _data
     ) external payable whenNotPaused {
-        // solhint-disable-next-line avoid-tx-origin
-        require(msg.sender == tx.origin, "Only EOA senders are allowed to send enforced transaction");
+        address sender = _msgSender();
+        if (sender != tx.origin) {
+            // alias smart contract account, consistent with message from L1ScrollMessenger.
```
Thegaram, line 98:
```suggestion
            // If sender is a classic SCA, then we apply address aliasing (consistent with message from L1ScrollMessenger).
            // If the sender is an EOA with no code, no aliasing is applied.
            // If the sender is an EIP-7702 delegated EOA, then aliasing behavior depends on which wallet initiated the call.
```

src/L1/gateways/EnforcedTxGateway.sol
Thegaram, line 1: Let's make `messageQueue` and `feeVault` `immutable`?

src/L1/rollup/ScrollChain.sol
jonastheis: I mean having the last `lastBatchHash` instead of `parentBatchHash` as parameter. This implies that we chose the correct `parentBatchHash` as well as that the batches have been generated as we expected. 

```diff
- function commitBatchesPostEuclidV2(uint8 version, bytes32 parentBatchHash)
+ function commitBatchesPostEuclidV2(uint8 version, bytes32 lastBatchHash)
// ...
+ parentBatchHash != committedBatches[lastCommittedBatchIndex]
// ...
- committedBatches[lastCommittedBatchIndex] = parentBatchHash;
+ if (parentBatchHash != lastBatchHash) revert InconsistentBatchHash(parentBatchHash, lastBatchHash);
+ committedBatches[lastCommittedBatchIndex] = parentBatchHash;
```

src/L1/rollup/IL1MessageQueueV2.sol
```diff
@@ -0,0 +1,114 @@
...
+    function firstCrossDomainMessageIndex() external view returns (uint256);
+
+    /// @notice The start index of all unfinalized messages.
+    function nextUnfinalizedQueueIndex() external view returns (uint256);
+
+    /// @notice Return the index of next appended message.
+    /// @dev Also the total number of appended messages, including messages in `L1MessageQueueV1`.
+    function nextCrossDomainMessageIndex() external view returns (uint256);
+
+    /// @notice Return the message rolling hash of in `queueIndex`.
```
Thegaram, line 44:
```suggestion
    /// @notice Return the message rolling hash of `queueIndex`.
```

src/L1/rollup/IL1MessageQueueV2.sol
```diff
@@ -0,0 +1,114 @@
...
+    /// @param queueIndex The index to query.
+    function getMessageRollingHash(uint256 queueIndex) external view returns (bytes32);
+
+    /// @notice Return the message enqueue timestamp of in `queueIndex`.
```
Thegaram, line 48:
```suggestion
    /// @notice Return the message enqueue timestamp of `queueIndex`.
```

src/L1/rollup/IL1MessageQueueV2.sol
Thegaram, line 1: What was the original rationale of introducing a new V2 contract? Did we consider simply upgrading the existing contract?

src/L1/rollup/IScrollChain.sol
```diff
@@ -41,6 +41,14 @@ interface IScrollChain {
     /// @param newMaxNumTxInChunk The new value of `maxNumTxInChunk`.
     event UpdateMaxNumTxInChunk(uint256 oldMaxNumTxInChunk, uint256 newMaxNumTxInChunk);
 
+    /// @notice Emitted when enter enforced batch mode.
+    /// @param lastCommittedBatchIndex The index of last committed batch.
+    event EnterEnforcedBatchMode(uint256 lastCommittedBatchIndex);
+
+    /// @notice Emitted when exit enforced batch mode.
+    /// @param lastCommittedBatchIndex The index of last committed batch.
+    event ExitEnforcedBatchMode(uint256 lastCommittedBatchIndex);
```
Thegaram, line 50: I'd probably just use a single event, but both ways are fine.

```suggestion
    /// @notice Emitted when we enter or exit enforced batch mode.
    /// @param enabled True if we are entering enforced batch mode, false otherwise.
    /// @param lastCommittedBatchIndex The index of last committed batch.
    event UpdateEnforcedBatchMode(bool enabled, uint256 lastCommittedBatchIndex);
```

src/L1/rollup/IScrollChain.sol
```diff
@@ -88,11 +96,15 @@ interface IScrollChain {
         bytes calldata blobDataProof
     ) external;
 
+    /// @notice Commit a batch after Euclid phase 2 upgrade.
+    /// @param version The version of current batch.
+    /// @param parentBatchHash The hash of parent batch.
+    function commitBatchesPostEuclidV2(uint8 version, bytes32 parentBatchHash) external;
```
Thegaram, line 102: Why not just call this `commitBatches`?

src/L1/rollup/IScrollChain.sol
```diff
@@ -88,11 +96,15 @@ interface IScrollChain {
...
+
     /// @notice Revert pending batches.
     /// @dev one can only revert unfinalized batches.
-    /// @param firstBatchHeader The header of first batch to revert, see the encoding in comments of `commitBatch`.
-    /// @param lastBatchHeader The header of last batch to revert, see the encoding in comments of `commitBatch`.
-    function revertBatch(bytes calldata firstBatchHeader, bytes calldata lastBatchHeader) external;
+    /// @param batchHeader The header of first batch to revert, see the encoding in comments of `commitBatch`.
```
Thegaram, line 106: This comment is inconsistent with the one at the implementation in `ScrollChain`.

Any specific reason why you decided to change the semantics from "first to remove" to "last to keep"?

src/L1/rollup/IScrollChain.sol
```diff
@@ -106,7 +118,41 @@ interface IScrollChain {
...
+        uint256 lastProcessedQueueIndex;
+        bytes32 postStateRoot;
+        bytes32 withdrawRoot;
+        bytes zkProof;
+    }
+
+    /// @notice Commit a batch of transactions on layer 1 and finalize it.
+    /// @param version The version of current batch.
+    /// @param finalizeStruct The data needed for finalize.
+    /// @param parentBatchHash The hash of parent batch.
```
Thegaram, line 152:
```suggestion
    /// @param parentBatchHash The hash of parent batch.
    /// @param finalizeStruct The data needed to finalize this batch.
```

src/L1/system-contract/SystemConfig.sol
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// @notice Parameters for enforced batch. The compiler will pack this struct into single `bytes32`.
+    /// @param maxDelayEnterEnforcedMode If no batch has been finalized for `maxDelayEnterEnforcedMode`,
+    ///        batch submission becomes permissionless. Anyone can submit a batch together with a proof.
+    /// @param maxDelayMessageQueue If message not included/finalized for `maxDelayMessageQueue`, the
+    ///        operator can not finalize any more batches until including those messages. If it continues
+    ///        to censor messages permissionless mode will eventually activate.
```
Thegaram, line 26: This is not consistent with the current implementation.

```suggestion
    /// @param maxDelayMessageQueue If no message is included/finalized for `maxDelayMessageQueue`,
    ///        batch submission becomes permissionless. Anyone can submit a batch together with a proof.
```

src/test/L1GatewayTestBase.t.sol
```diff
@@ -96,37 +102,76 @@ abstract contract L1GatewayTestBase is ScrollTestBase {
...
+                    1233,
+                    address(messageQueueV1),
+                    address(messageQueueV2),
+                    address(verifier),
+                    address(system)
+                )
+            )
         );
-        rollup.initialize(address(messageQueue), address(0), 44);
+        rollup.initialize(address(messageQueueV1), address(0), 44);
```
Thegaram, line 174: Do we need to call `initializeV2`?

src/L1/rollup/ScrollChain.sol
```diff
@@ -239,13 +273,44 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         emit UpdateMaxNumTxInChunk(0, _maxNumTxInChunk);
     }
 
+    function initializeV2() external reinitializer(2) {
+        // binary search on lastCommittedBatchIndex
```
Thegaram, line 277: Would be good to have a fuzz test for this

src/L1/rollup/ScrollChain.sol
```diff
@@ -328,45 +395,39 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+        for (uint256 _batchIndex = lastBatchIndex; _batchIndex > startBatchIndex; --_batchIndex) {
             bytes32 _batchHash = committedBatches[_batchIndex];
             committedBatches[_batchIndex] = bytes32(0);
 
-            // also revert initial Euclid batch
-            if (_initialEuclidBatchIndex == _batchIndex) {
-                initialEuclidBatchIndex = 0;
-            }
-
             emit RevertBatch(_batchIndex, _batchHash);
```
Thegaram, line 426: > during commit batch, we only store last batch hash into storage

Doesn't that mean that we will emit lots of `RevertBatch(index, 0x0)` events? This significantly alters the semantics of this event.

src/L1/rollup/ScrollChain.sol
```diff
@@ -328,45 +395,39 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
 
-        // `getL1MessagePopped` codes are the same in V0~V6
-        uint256 l1MessagePoppedFirstBatch = BatchHeaderV0Codec.getL1MessagePopped(firstBatchPtr);
-        unchecked {
-            IL1MessageQueue(messageQueue).resetPoppedCrossDomainMessage(
-                _totalL1MessagesPoppedOverallFirstBatch - l1MessagePoppedFirstBatch
-            );
-        }
+        // update `lastCommittedBatchIndex`
+        miscData.lastCommittedBatchIndex = uint64(startBatchIndex);
```
Thegaram, line 430: We need to ensure that `lastCommittedBatchIndex` has a batch index in storage. Or is that already ensured by `_loadBatchHeader`?

src/L1/rollup/ScrollChain.sol
```diff
@@ -408,28 +469,63 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+    }
+
+    /// @inheritdoc IScrollChain
+    /// @dev We only consider batch version >= 7 here.
+    function commitAndFinalizeBatch(
+        uint8 version,
+        bytes32 parentBatchHash,
+        FinalizeStruct calldata finalizeStruct
+    ) external {
+        // we use BatchHeaderV7Codec for finalizeStruct.batchHeader
```
Thegaram, line 493: This comment seems to be out of place here

src/L1/rollup/ScrollChain.sol
```diff
@@ -408,28 +469,63 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+            uint256 lastUnfinalizedMessageTime = IL1MessageQueueV2(messageQueueV2)
+                .getFirstUnfinalizedMessageEnqueueTime();
+            if (
+                lastUnfinalizedMessageTime + maxDelayMessageQueue < block.timestamp ||
+                cachedMiscData.lastFinalizeTimestamp + maxDelayEnterEnforcedMode < block.timestamp
+            ) {
+                // explicit set enforce batch enable
+                cachedMiscData.enforcedModeEnabled = true;
+                // reset `lastCommittedBatchIndex`
+                cachedMiscData.lastCommittedBatchIndex = uint64(miscData.lastFinalizedBatchIndex);
```
Thegaram, line 507: Any concern about stale data in `committedBatches`?

src/L1/rollup/ScrollChain.sol
```diff
@@ -494,17 +590,30 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         }
     }
 
+    /// @notice Exit from enforced batch mode.
+    function disableEnforcedBatch() external onlyOwner {
```
Thegaram, line 594:
```suggestion
    function disableEnforcedBatchMode() external onlyOwner {
```

src/L1/rollup/ScrollChain.sol
```diff
@@ -650,6 +775,100 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
+            BatchHeaderV0Codec.storeBatchIndex(batchPtr, lastCommittedBatchIndex);
+            BatchHeaderV7Codec.storeParentBatchHash(batchPtr, parentBatchHash);
+            BatchHeaderV7Codec.storeBlobVersionedHash(batchPtr, blobVersionedHash);
+            bytes32 batchHash = BatchHeaderV0Codec.computeBatchHash(
+                batchPtr,
+                BatchHeaderV7Codec.BATCH_HEADER_FIXED_LENGTH
+            );
+            emit CommitBatch(lastCommittedBatchIndex, batchHash);
+            parentBatchHash = batchHash;
+            if (onlyOne) break;
```
Thegaram, line 816: Do we need to check that there are no other blobs?

src/L1/gateways/EnforcedTxGateway.sol
zimpha: done

src/L1/rollup/IL1MessageQueueV2.sol
zimpha: use a new V2 contract would be more clean. Our v1 contract has some used features like base fee oracle updates.

src/L1/rollup/ScrollChain.sol
zimpha: we have one for this, in `ScrollChainTest.testInitializeV2`.

src/L1/rollup/ScrollChain.sol
zimpha: yes, it is ensured in `_loadBatchHeader`.

src/L1/rollup/ScrollChain.sol
zimpha: I don't see any problems. We now use `lastCommittedBatchIndex` to determine whether a batch is committed or now.

src/L1/rollup/ScrollChain.sol
zimpha: no need, since we use `CommitBatch` event to match the corresponding blob.

src/test/L1GatewayTestBase.t.sol
zimpha: no need, but I add the `initializeV2` call here.

src/test/ScrollChain.t.sol
```diff
@@ -334,540 +502,395 @@ contract ScrollChainTest is DSTestPlus {
...
+        // import 10 L1 messages into message queue V2
+        for (uint256 i = 0; i < 10; i++) {
+            messageQueueV2.appendCrossDomainMessage(address(this), 1000000, new bytes(0));
+        }
+        assertEq(messageQueueV2.nextUnfinalizedQueueIndex(), 10);
+        assertEq(messageQueueV2.nextCrossDomainMessageIndex(), 20);
+        assertEq(messageQueueV2.getFirstUnfinalizedMessageEnqueueTime(), 100);
+        hevm.warp(100 + 86400 + 1);
+
+        // succeed to call commitAndFinalizeBatch 13
```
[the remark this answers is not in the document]
zimpha: added.

src/L1/rollup/ScrollChain.sol
zimpha: changed to @jonastheis's way.

src/L1/rollup/ScrollChain.sol
zimpha: added

src/L1/rollup/IScrollChain.sol
zimpha: we need to call `_loadBatchHeader` in `revertBatch`, during the call it will check `keccak256(batchHeader)` with `committedBatches`. For intermediate batches, the value of `committedBatches` is `bytes32(0)`, the check will always failed.

With the "last to keep" semantics, we can make sure the given batchHeader has corresponding nonzero `committedBatches`.

src/L1/rollup/ScrollChain.sol
zimpha: I changed the event to only include the range now. So the zero batch hash won't be a problem. But we need to inform third-party about this, since they may need to watch the events.

src/L1/rollup/ScrollChain.sol
zimpha: removed

src/L1/gateways/EnforcedTxGateway.sol
```diff
@@ -53,18 +63,18 @@ contract EnforcedTxGateway is OwnableUpgradeable, ReentrancyGuardUpgradeable, Pa
...
-    constructor() {
+    constructor(address _messageQueue, address _feeVault) {
         _disableInitializers();
+
+        messageQueue = _messageQueue;
+        feeVault = _feeVault;
     }
 
-    function initialize(address _queue, address _feeVault) external initializer {
+    function initialize(address, address) external initializer {
```
Thegaram, line 73: Why not just remove the arguments?

src/L1/rollup/IScrollChain.sol
```diff
@@ -151,12 +153,7 @@ interface IScrollChain {
...
     /// @param version The version of current batch.
-    /// @param parentBatchHash The hash of parent batch.
     /// @param finalizeStruct The data needed to finalize this batch.
     /// @dev The batch payload is stored in the blob.
-    function commitAndFinalizeBatch(
-        uint8 version,
-        bytes32 parentBatchHash,
-        FinalizeStruct calldata finalizeStruct
-    ) external;
+    function commitAndFinalizeBatch(uint8 version, FinalizeStruct calldata finalizeStruct) external;
```
Thegaram, line 158: Is the idea here that if the assumed `parentBatchHash` (or `lastBatchHash`) is incorrect (e.g. if someone else committed a batch in a meantime), then the proof verification will fail anyway?

src/L1/rollup/ScrollChain.sol
```diff
@@ -427,12 +440,7 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
 
         // actual revert
         uint256 lastBatchIndex = miscData.lastCommittedBatchIndex;
-        for (uint256 _batchIndex = lastBatchIndex; _batchIndex > startBatchIndex; --_batchIndex) {
-            bytes32 _batchHash = committedBatches[_batchIndex];
-            committedBatches[_batchIndex] = bytes32(0);
-
-            emit RevertBatch(_batchIndex, _batchHash);
-        }
+        emit RevertBatch(startBatchIndex + 1, lastBatchIndex);
```
Thegaram, line 443: Check that `startBatchIndex < lastBatchIndex`.

src/L1/rollup/ScrollChain.sol
```diff
@@ -488,19 +496,26 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
-    function commitAndFinalizeBatch(
-        uint8 version,
-        bytes32 parentBatchHash,
-        FinalizeStruct calldata finalizeStruct
-    ) external {
-        // we use BatchHeaderV7Codec for finalizeStruct.batchHeader
+    function commitAndFinalizeBatch(uint8 version, FinalizeStruct calldata finalizeStruct) external {
         ScrollChainMiscData memory cachedMiscData = miscData;
-        if (!cachedMiscData.enforcedModeEnabled) {
+        if (!_decodeBoolFromFlag(cachedMiscData.flags, ENFORCED_MODE_OFFSET)) {
```
Thegaram, line 518: `!isEnforcedModeEnabled()`

src/L1/rollup/ScrollChain.sol
```diff
@@ -816,39 +852,35 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
             if (onlyOne) break;
         }
 
+        if (parentBatchHash != lastBatchHash) {
```
jonastheis, line 855: Would be great to add a comment here why and how this could happen:
- faulty batch producer
- unexpected `parentBatch` in case commit transactions get reordered 
- anything else?

src/L1/rollup/ScrollChain.sol
```diff
@@ -480,19 +496,26 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         bytes32 withdrawRoot,
         bytes calldata aggrProof
     ) external override OnlyProver whenNotPaused whenEnforcedBatchNotEnabled {
+        uint256 flags = miscData.flags;
```
jonastheis, line 499: Why don't we check this in `_finalizeBundlePostEuclidV2`? Do we not need to check this for `commitAndFinalizeBatch`?

src/L1/rollup/IScrollChain.sol
jonastheis: I guess that bc we're only committing a single batch for which we need to specify the batch header we can just hash it and pass it as the `lastBatchHash` to the commit function

zimpha: @jonastheis is correct, only 1 batch is committed. If someone else committed a batch in a meantime, it will revert during committing.

src/L1/rollup/ScrollChain.sol
zimpha: It is unlikely that our sequencer/prover is down during the upgrade. So when someone call `commitAndFinalizeBatch`, all v1 messages are finalized.

Thegaram: Added in a6d76274ef6d9df1298bd92bd3f5cdf64fd9c5aa

src/L1/rollup/ScrollChain.sol
```diff
@@ -427,12 +440,7 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
 
         // actual revert
         uint256 lastBatchIndex = miscData.lastCommittedBatchIndex;
-        for (uint256 _batchIndex = lastBatchIndex; _batchIndex > startBatchIndex; --_batchIndex) {
-            bytes32 _batchHash = committedBatches[_batchIndex];
-            committedBatches[_batchIndex] = bytes32(0);
```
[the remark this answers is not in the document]
Thegaram: In hindsight I'd rather add this to `_loadBatchHeader`, easier to avoid incorrect usage. But that's one extra `SLOAD` for `misc`. Your call.

src/L1/rollup/ScrollChain.sol
```diff
@@ -515,17 +666,22 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
     {
         // check whether the batch is empty
         if (_chunks.length == 0) revert ErrorBatchIsEmpty();
-        (, _parentBatchHash, _batchIndex, _totalL1MessagesPoppedOverall) = _loadBatchHeader(_parentBatchHeader);
+        uint256 batchPtr;
+        (batchPtr, _parentBatchHash, _batchIndex, _totalL1MessagesPoppedOverall) = _loadBatchHeader(_parentBatchHeader);
+        // version should non-decreasing
+        if (BatchHeaderV0Codec.getVersion(batchPtr) > _version) revert ErrorCannotDowngradeVersion();
+
+        if (_batchIndex != _lastCommittedBatchIndex) revert ErrorBatchIsAlreadyCommitted();
```
Thegaram, line 682: Note:
- `_batchIndex < _lastCommittedBatchIndex` => batch already committed
- `_batchIndex > _lastCommittedBatchIndex` => batch reverted

Thegaram (approved):

src/L1/rollup/ScrollChain.sol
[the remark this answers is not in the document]
zimpha: done in https://github.com/scroll-tech/scroll-contracts/pull/74/commits/c0d6c658684aeabfc10357956a0e4d2e66b310e7

reorder some statement so no extra `SLOAD` is needed.

src/L1/rollup/ScrollChain.sol
```diff
@@ -561,15 +718,17 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
         uint256 totalL1MessagesPoppedOverall,
         bytes32 postStateRoot,
-        bytes32 withdrawRoot
+        bytes32 withdrawRoot,
+        bool isV1
     ) internal {
         // @note we do not store intermediate finalized roots
-        lastFinalizedBatchIndex = batchIndex;
+        miscData.lastFinalizedBatchIndex = uint64(batchIndex);
+        miscData.lastFinalizeTimestamp = uint32(block.timestamp);
```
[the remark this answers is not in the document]
zimpha: done in https://github.com/scroll-tech/scroll-contracts/pull/74/commits/c0d6c658684aeabfc10357956a0e4d2e66b310e7

src/L1/rollup/ScrollChain.sol
```diff
@@ -328,45 +428,33 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
-            committedBatches[_batchIndex] = bytes32(0);
-
-            // also revert initial Euclid batch
-            if (_initialEuclidBatchIndex == _batchIndex) {
-                initialEuclidBatchIndex = 0;
-            }
-
-            emit RevertBatch(_batchIndex, _batchHash);
-        }
+        emit RevertBatch(startBatchIndex + 1, lastBatchIndex);
```
Thegaram, line 454: If `startBatchIndex == lastBatchIndex` then this event will be a bit weird. But it's probably OK, we can treat any event where `startBatchIndex > finishBatchIndex` as a no-op.

Thegaram (approved):

Thegaram (approved):

src/L1/rollup/ScrollChain.sol
```diff
@@ -163,21 +199,31 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
...
-    /// @inheritdoc IScrollChain
-    uint256 public override lastFinalizedBatchIndex;
+    /// @dev The storage slot used as `lastFinalizedBatchIndex`, which is deprecated now.
+    uint256 private __lastFinalizedBatchIndex;
 
     /// @inheritdoc IScrollChain
+    /// @dev Starting from EuclidV2, this array is sparse: it only contains
+    /// the last batch hash per commit transaction, and not intermediate ones.
+    /// @dev Starting from EuclidV2, this array might contain reverted batches,
+    /// callers must check that `batchIndex <= miscData.lastCommittedBatchIndex`.
```
Thegaram, line 209: If this comment is invalidated by [e969619](https://github.com/scroll-tech/scroll-contracts/pull/74/commits/e96961987369e7655730fdbe9338206230e41645), then please remove or update it.

zimpha: removed in https://github.com/scroll-tech/scroll-contracts/pull/74/commits/de8fa12266bd71e69fddc185c7ee49bb9001289c

PR #144, opened 2025-09-15
Validium verifier with (domain,version) and updated public input

roynalnaruto:
- update comment for public inputs (appending 33-bytes for encryption key)
- include the new `(domain, stf_version)` structure for version byte, where `version_byte = (domain << 6) + stf_version`


roynalnaruto:
Closing as we can pass `protocolVersion = (domain << 6) + stf_version` at the time of contract (`ZkEvmVerifierPostFeynman.sol`) deployment.

